PR #17, opened 2017-07-14
Maestro Woof!

gonsie:
I created a logo for your project! For some reason the name just looks like "woof" to me. So the logo is a dog with a conductor's baton.


IanLee1521:
Maybe better in line with the title, rather than below?

```
# Maestro Workflow Conductor (MaestroWF) <img src="docs/images/logo.png" alt="MaestroWoof" height="100px"/>
```

Issue #9, opened 2017-07-13
Add the ability to split allocations


FrankD412:
Currently a workflow step is the only granularity for allocating nodes. There is not syntax to break apart a single step's allocation into further sub-allocations using a parallelization command. This lack of functionality prevents users from running multiple commands that may need to be run concurrently -- currently such commands would have to be scheduled independently limiting the possible flows that can be represented.

FrankD412:
A new branch exists for this feature -- Branch 1296f99eab48a90de4c3f332c90ed0e1a28fa236

FrankD412:
This was implemented by PR #19. Closing.

PR #19, opened 2017-07-18
Alloc

FrankD412:
Address the functionality discussed in Issue #9. This PR adds the ability to split allocations within a scheduler using the "$(LAUNCHER)[[nodes],[procs]]" nomenclature.


FrankD412:
It's come to my attention that there's another parameter that can dictate allocation -- tasks (for SLURM at least).  This PR will just handle adding nodes and processors.

PR #24, opened 2017-07-18
Add ci supportfix

FrankD412:
Addresses the issue of support Python versions and updates tox settings to appropriately mirror the selection of Python 2.7 and 3.6.


.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: false
+language: python
+python:
+    - "2.6"
```
IanLee1521: This list should match what is in the tox.ini file (you probably can do this entire list, but you may need to fine tune it.

IanLee1521: And setup.py

FrankD412: I fixed .travis.yaml -- but for setup.py are you referring to using the ```python_requires``` entry?

setup.py
```diff
@@ -23,9 +23,9 @@
       classifiers=[
         'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
-        'Operating System :: OS Independent',
+        'Operating System :: Unix',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 2.7',
```
IanLee1521: These classifiers should be updated to include all relevant python versions.

PR #26, opened 2017-07-18
Updated travis.yml to include python 3.4 and 3.5

IanLee1521:
Signed-off-by: Ian Lee <lee1001@llnl.gov>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,13 @@
-sudo: false
 language: python
+
+python:
     - "2.7"
+    - "3.4"
+    - "3.5"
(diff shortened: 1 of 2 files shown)


FrankD412:
So looking at the TravisCI errors, states having permission issues when installing tox-travis. Does tox-travis need to be installed in the user level using ```pip --user```?

IanLee1521:
@FrankD412 -- Should be good to go, finally able to poke this along.

FrankD412:
Oh, I see -- It goes by the qualifiers. Good thing to know. Merging.

PR #31, opened 2017-07-26
Merging of multiple piecewise specifications

FrankD412:
Introduces the ability to take multiple YAMLSpecifications be merged together via an overridden addition operator.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -157,6 +157,10 @@ def main():
         spec += item
 
     spec.verify()
+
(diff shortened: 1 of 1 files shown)


FrankD412:
Removing from the latest milestone -- This PR turned out to be a lot more complicated than initially anticipated.

FrankD412:
Closing this PR since there is still on-going discussion and I have other thoughts on how to implement this which would be more extensive than this PR originally entailed.

PR #38, opened 2017-08-02
Addition of a LocalScriptAdapter to execute workflows on a user's local machine

FrankD412:


FrankD412:
Ok – so this one is actually much more complicated than I realized. There are two ways to achieve this:
*The quick way which is to just lift the requirement that the YAML specification have a batch entry in it. The only real change this method prompts is to make sure that the study steps do not have procs and nodes specified (because those entries in steps now becomes invalid). Since the ExecutionGraph detects when procs and nodes are not specified it will automatically execute locally. There is a major drawback to this method though – the ExecutionGraph executes steps sequentially and will wait for a process to terminate before moving onto the next step (the local execution in the ExecutionGraph was coded with small steps running locally in mind).
*The more extensive way is as follows – using the multiprocessing library that's standard to Python this scheme would implement a LocalScriptAdapter which contains a process Pool object within it that would have a set number of processes reading from a single queue. The ExecutionGraph would submit using this adapter just like it would any other adapter with some hardcoding to know not to check for procs and nodes. The drawback here is that the Pool object is not able to be pickled – this means that there would have to be decent overhaul of how an adapter is stored within the ExecutionGraph. A couple of changes are needed to accomplish this overhaul:
The ExecutionGraph can no longer store the adapter instance itself. Because the LocalScriptAdapter would in principle emulate a scheduler to the graph, that would mean the Pool instance and Queue would live in adapter itself. That prevents the adapter from being pickled (the Pool and Queue cannot be pickled). That leaves us with limited options – primarily this would necessitate making the adapter factory and storing just the adapter type string in the ExecutionGraph so that it can be instantiated at execution time.
This requires more extensive implementation and testing since we're not using multiprocessing currently.

FrankD412:
Closing this PR because I revisited the implementation. The new PR is #41

Issue #29, opened 2017-07-19
Formalize a PyPi Release Process


FrankD412:
We need to formalize a release process to PyPi that will automatically compile all supported versions for both sdist and bdist.

IanLee1521:
Would you be willing to make me a POC for the project in PYPI?

Same username "IanLee1521"

FrankD412:
I've added you as a maintainer, did it go through?

FrankD412:
@IanLee1521 -- Alright looks like you are a maintainer. Is there an accepted way of formalizing release. It looks like tagging, and using something like this article?

http://www.dougalmatthews.com/2016/Sep/01/automate-publishing-to-pypi-with-pbr-and-travis/

Do you know if that's the most current accepted way?

IanLee1521:
That's a bit fancier than what I've done... It tends to be (for most projects):

- Make a commit with the updated version string: ``git commit -am'Released version X.Y'``
- Tag that commit: ``git tag -m 'Released version X.Y' X.Y``
- Run build and release target, something like: ``python setup.py sdist bdist_wheel && twine upload dist/*``

I tend to do something like have a ``Makefile`` target that does the right thing, at least for the building and pushing part (I prefer to do the first two steps manually).

FrankD412:
Got it -- Might it be worthwhile exploring something fancier? It'd imagine that the CI integration will halt a release if it finds an issue.

IanLee1521:
I guess the question is "how often are you planning to do a release?" and what branches etc might you do that on?

Yes you could automate, but it might make sense to hold off until there is something to automate ;)

IanLee1521:
Also, I'm not positive how / if you could automate the pushing to PYPI... that might hold you up... (because you need to authenticate to the server).

FrankD412:
Fair enough. I guess the better way to do this would be to just have a shell script (or maybe a make clause that takes a version?) that you give it the version and it automatically executes and then just lets you put in your password.

IanLee1521:
That’s about as far as I would go with it, yes.

PR #39, opened 2017-08-04
First pass at a release shell script.

FrankD412:
Addition of an automated release process shell script.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+git commit -am "Released version $1"
+git tag -m "Released version $1" $1
+git push origin $1
+
+python setup.py sdist bdist_wheel
+
+twine upload dist/*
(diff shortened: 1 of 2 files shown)


FrankD412:
@IanLee1521 - I figure that if you okay the shell script we'll go ahead and merge it. I'm not sure how to test this until we're ready for a release.

FrankD412:
This addresses issue #29

release.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+git commit -am "Released version $1"
+git tag -m "Released version $1" $1
+git push origin $1
+
+python setup.py sdist bdist_wheel
```
IanLee1521: Instead of building two different wheels, if your code is python 2/3 compatible, you should add a `setup.cfg` file and define the project as a universal wheel. Then call bdist_wheel just the one time. See: https://packaging.python.org/tutorials/distributing-packages/#universal-wheels

Issue #49, opened 2017-09-15
Crash Log Traceback stuck at localscriptadapter.py


faical-yannick-congo:
When i run a study and a step fails i have in the stderr trace in the .txt file but it is stuck at localscriptadapter.py. Here is a trace for SEM-Images-Stats.txt.
------
Traceback (most recent call last):
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file localscriptadapter.py, line 138
Traceback (most recent call last):
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file localscriptadapter.py, line 138
Traceback (most recent call last):
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/Users/fyc/anaconda2/envs/demo-2.7/lib/python2.7/logging/__init__.py", line 329, in getMessage
[...]

PR #57, opened 2017-10-31
Correct logger.warning call in localscriptadapter

dinatale2:
logger.warning uses % to perform string formatting. % does
not support new style string formatting with {}.


FrankD412:
This is meant to correct the bug in #49.

FrankD412:
So, after testing this PR, it does fix the exception; however, the way it's structured is that it pipes stderr to the log. Ultimately, the error being properly reported all relies on an application using stderr to report its errors instead of stdout. I've already encountered cases where stdout is used.

PR #66, opened 2018-01-23
Enhancement/submission throttle

FrankD412:
This PR adds the ability to set a maximum number of in progress jobs.

### DIFF
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -4,6 +4,7 @@ import getpass
 import logging
 import os
 import pickle
+from collections import deque
 
 from maestrowf.abstracts.enums import JobStatusCode, State, SubmissionCode, \
     CancelCode
@@ -215,12 +216,14 @@ class ExecutionGraph(DAG):
     where that would go.
     """
 
-    def __init__(self, submission_attempts=1):
+    def __init__(self, submission_attempts=1, submission_throttle=0):
         """
         Initialize a new instance of an ExecutionGraph.
 
         :param submission_attempts: Number of attempted submissions before
         marking a step as failed.
+        :param submission_throttle: Maximum number of scheduled in progress
+        submissions.
         """
         super(ExecutionGraph, self).__init__()
         # Member variables for execution.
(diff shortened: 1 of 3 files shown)


maestrowf/datastructures/core/executiongraph.py
```diff
@@ -642,22 +663,37 @@ def execute_ready_steps(self):
...
-                record.name, record.status
-            )
-            self._execute_record(key, record)
-            logger.debug(
-                "After execution of '%s' -- New state is %s.",
-                record.name, record.status
-            )
+        # If we don't have a submission limit, go ahead and submit all.
+        if not self._submission_throttle:
+            while self.ready_steps:
```
dinatale2: There is no need to have two loops here. Your `if else` can set the value for `_available` and you can reuse a loop.

maestrowf/maestro.py
```diff
@@ -105,7 +105,7 @@ def run_study(args):
     # Setup the study.
     study = Study(spec.name, spec.description, studyenv=environment,
                   parameters=parameters, steps=steps)
-    study.setup()
+    study.setup(throttle=args.throttle)
```
dinatale2: You're using `args.throttle` without confirming that it is `0` or greater.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -642,22 +663,30 @@ def execute_ready_steps(self):
...
-                "Attempting to execute '%s' -- Current state is %s.",
-                record.name, record.status
-            )
-            self._execute_record(key, record)
-            logger.debug(
-                "After execution of '%s' -- New state is %s.",
-                record.name, record.status
-            )
+        # If we don't have a submission limit, go ahead and submit all.
+        if not self._submission_throttle:
```
dinatale2: Doesn't this test if `self._submission_throttle` is not `None`? Currently, `self._submission_throttle` can't be `None` right? Since the default is set to `0` on `__init__`.

maestrowf/maestro.py
```diff
@@ -105,7 +105,14 @@ def run_study(args):
...
     study = Study(spec.name, spec.description, studyenv=environment,
                   parameters=parameters, steps=steps)
-    study.setup()
+
+    # Check if the throttle is zero or greater:
+    if args.throttle < 0:
+        _msg = "Submission throttle must be a value of zero. '{}' provided." \
+            .format(args.throttle)
+        LOGGER.error(_msg)
+        raise ArgumentError(_msg)
```
dinatale2: nit: This would benefit from a line of whitespace below it.

maestrowf/datastructures/core/executiongraph.py
FrankD412: This works because the value of 0 is considered to be a False value.

maestrowf/maestro.py
```diff
@@ -105,7 +105,14 @@ def run_study(args):
     # Setup the study.
     study = Study(spec.name, spec.description, studyenv=environment,
                   parameters=parameters, steps=steps)
-    study.setup()
+
+    # Check if the throttle is zero or greater:
+    if args.throttle < 0:
+        _msg = "Submission throttle must be a value of zero. '{}' provided." \
```
dinatale2: `Submission throttle must be a value of 0 or greater. '{}' provided.`?

PR #73, opened 2018-02-03
[TEST] Run local lulesh example

dinatale2:
Testing if I can get `lulesh` to run via `tox`.

### DIFF
--- a/scripts/run_lulesh.sh
+++ b/scripts/run_lulesh.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


FrankD412:
Just as a note, this was put up on Travis and runs; however, there is no event or anything that if you launch the conductor via nohup using Maestro. A way around this is to use the Maestro front end to construct the study scripts and then call the conductor manually (since the conductor only returns once the study is completed).

FrankD412:
@dinatale2 -- I'm taking over this PR if you don't mind.

PR #75, opened 2018-02-09
Adds support for enhanced launcher token notations.

### LINKED ISSUE #30: Extend Sub-allocation Functionality to support task granularity
The current implementation of the sub-allocation feature only supports assigning with a granularity of nodes and processors. Tasks are also another granularity supported by some (maybe most?) schedulers and so it should be supported.

FrankD412:
This PR continues to support the legacy notation of $(LAUNCHER)[n,p],
where n is the number of nodes a command requires and p is the number
of processors. This notation is expanded to loosen the requirement for
nodes (except in the legacy notation above) and changes the notation to
allow for more flexibility (nodes and processors can be notated out of
the above order). The new notations becomes $(LAUNCHER)[#n, #p] where
'#n' is not required ($(LAUNCHER)[#p] is also valid). See the study in
samples/documentation/launcher_tokens.yaml. Closes #30.

Initial logic and new regex.

Logic for parsing out launcher tokens.

Addition of a case for when launcher token appears on its own.

Update to the SLURM adapter to supposed unset node counts.

Added launcher token documentation study file.

Bug fixes for regex matching and processing.

Signed-off-by: Francesco Di Natale <dinatale3@llnl.gov>

### DIFF
--- a/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
+++ b/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
@@ -51,10 +51,17 @@ class SchedulerScriptAdapter(ScriptAdapter):
     # The var tag to look for to replace for parallelized commands.
     launcher_var = "$(LAUNCHER)"
     # Allocation regex and compilation
-    alloc_regex = re.compile(
-        re.escape(launcher_var) +
-        r"\[(?P<nodes>[0-9]+),\s*(?P<procs>[0-9]+)\]"
-    )
+    # Keeping this one here for legacy.
+    launcher_regex = re.compile(
+        re.escape(launcher_var) + r"\[(?P<alloc>.*)\]")
+
+    # We can have multiple requested submission properties.
+    # Legacy allocation of nodes and procs.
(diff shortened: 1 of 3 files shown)


maestrowf/abstracts/interfaces/schedulerscriptadapter.py
```diff
@@ -112,62 +119,102 @@ def _substitute_parallel_command(self, step_cmd, nodes, procs):
...
-                    raise ValueError(msg)
-                if int(alloc_procs) > procs:
-                    msg = err_msg.format(
-                        match.group(),
-                        alloc_procs,
-                        "procs",
-                        procs
+                else:
+                    # We're dealing with the new style.
+                    _nodes = re.search(self.node_alloc, _alloc)
```
dinatale2: I think this allows you to have malformed allocation breakdown. The way I'm understanding the code, you would allow something like `[3n,3p,2n]` and you'd end up with 3 nodes and 3 processes currently. But, that's clearly wrong. This may imply that there is a need for a validation subcommand.

PR #80, opened 2018-03-13
Adds the ability to specify reservations and other scheduling parameters.

FrankD412:
This PR was originally meant to just add reservations, but the abstract interfaces needed to be updated to support it. There were updates and some clean up to the abstract `SchedulerScriptAdapter` and `ScriptAdapter`. The `SchedulerScriptAdapter` now passes a `**kwargs` parameter to support a more flexible range of parameters.

NOTE: This PR is a precursor to a wider change to how resources will be specified in the specification per step.

### DIFF
--- a/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
+++ b/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
@@ -94,7 +94,7 @@ class SchedulerScriptAdapter(ScriptAdapter):
         pass
 
     @abstractmethod
-    def get_parallelize_command(self, nodes, proc, **kwargs):
+    def get_parallelize_command(self, procs, nodes, **kwargs):
         """
         Generate the parallelization segement of the command line.
 
--- a/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
+++ b/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
@@ -94,7 +94,7 @@ class SchedulerScriptAdapter(ScriptAdapter):
         pass
 
     @abstractmethod
-    def get_parallelize_command(self, **kwargs):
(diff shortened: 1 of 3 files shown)


maestrowf/abstracts/interfaces/schedulerscriptadapter.py
```diff
@@ -95,7 +94,7 @@ def get_header(self, step):
         pass
 
     @abstractmethod
-    def get_parallelize_command(self, procs, nodes=None):
+    def get_parallelize_command(self, nodes, proc, **kwargs):
```
dinatale2: You swapped the order of the arguments `nodes` and `proc`, but you use them in the original order below.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -149,7 +157,18 @@ def submit(self, step, path, cwd, job_map=None, env=None):
         :returns: The return status of the submission command and job
         identiifer.
         """
-        cmd = " ".join(["sbatch", path, "-D", cwd])
+        # Check and see if we should be submitting into a reservation.
+        if self._batch["reservation"]:
```
dinatale2: style: I feel like this could be cleaned up. Build a collection for command? Something like

```
args = ["sbatch"]

if self._batch["reservation"]:
    args += ["--reservation", self._batch["reservation"]]

args += ["-D", cwd, path]
cmd = " ".join(args)
```

maestrowf/abstracts/interfaces/schedulerscriptadapter.py
FrankD412: That was unintentional. Fixed.

FrankD412:
Requested changes made.

PR #82, opened 2018-03-23
Flux adapter (Spectrum MPI)

trws:


maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -151,22 +152,28 @@ def get_parallelize_command(self, procs, nodes=None, **kwargs):
...
+        #     '-mca', 'plm', 'rsh',
+        #     # '-mca', 'plm_rsh_agent', 'rsh',
+        #     '--map-by', 'node']
+        # args.extend(['-hostfile','$HOSTF'])
+        # args.extend([
+        #     '-n',
+        #     str(procs),
+        #     # '/usr/gapps/kras/install/bin/autobind', # bind cores and GPUs
+        #     ])
+        args = ['/usr/gapps/kras/install/bin/mpiexec.hydra']
```
FrankD412: This looks platform specific -- we should probably assume mpiexec is somewhere in the path.

trws:
Much as I agree with you in general, the whole thing is currently 
platform specific.  That command shouldn’t really be used with flux 
anyway, it should be wreckrun doing this part, but this is faster and 
probably more stable to try and get splash working.  I can change it to 
just “mpiexec.hydra” but it will still be platform specific.

FrankD412:
I'm okay with that. Mostly just something to avoid the path.

trws:
If you want to, that would work, but the `mpiexec.hydra` thing is not right for either flux normally or for spectrum.   It should be mpirun for spectrum or `flux wreckrun` for flux.  If you want to merge this, put up an issue for that first so we make sure to address it.

FrankD412:
What's the proper mpirun for spectrum? We're still aiming to make this adapter fairly platform specific aren't we? Or are we thinking of launching flux in a more general way where we can just avoid Spectrum MPI entirely (since we had decided to preamble a general adapter with the Spectrum variant)?

trws:
mpirun with all the extra flags is what we have to use for spectrum.  
Yeah, this one is still platform specific, but hydra only works to 
launch mpich apps or flux itself, it can’t bootstrap spectrum.

FrankD412:
Do you mind making the quick correction to the command so that it can work and then pushing it? I'll go ahead and merge it after that.

trws:
Sure, just a sec.

trws:
That should do it, we should run some tests after this to make sure nothing in splash got accidentally broken though.

trws:
Actually, given the new drop of *all* the software on there, we probably should anyway.

FrankD412:
I agree -- we could merge this and apply fixes as needed (because I expect we're gonna be changing a lot of other things anyway)

trws:
Agreed.

PR #83, opened 2018-04-03
 Allow for list of labels in parameter

koning:
If the parameter in the global.parameters section of the yaml file is
not a simple number, the resulting label will give non-viable directory
names.

### DIFF
--- a/maestrowf/interfaces/script/slurmscriptadapter.py
+++ b/maestrowf/interfaces/script/slurmscriptadapter.py
@@ -110,7 +110,8 @@ class SlurmScriptAdapter(SchedulerScriptAdapter):
             # If we're looking at the bank and the reservation header exists,
             # skip the bank to prefer the reservation.
             if key == "bank" and "reservation" in self._batch:
-                continue
+                if self._batch["reservation"]:
+                    continue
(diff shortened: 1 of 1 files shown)


FrankD412:
@koning -- A couple of things.

1. `flake8` is failing -- it states that bare `except` statements should not be used. Can you rework the logic? I think the simplest way to accomplish this is to check if we have an instance of a `str` or `list`.
2. You do not appear to be checking that the length of the label list is the same as the length as the length of the number of parameters. Currently, the only supported way to define parameters is to enumerate out all combinations. The `YAMLSpecification` class has verification methods that could handle this.

maestrowf/datastructures/core/parameters.py
```diff
@@ -299,7 +299,10 @@ def get_combinations(self):
             combo = Combination()
             for key in self.parameters.keys():
                 pvalue = self.parameters[key][i]
-                tlabel = self.labels[key].replace(self.label_token,
+                try:
+                    tlabel = self.labels[key][i]
+                except:
```
FrankD412: `flake8` flags bare `except` statements as a style violation. I suggest explicitly checking for types here.

PR #84, opened 2018-04-04
 Allow for list of labels in parameter

koning:
THis PR adds the ability to define the label for a global parameter as a unique list of label names. The yamlspecification will now check for list types and check for length agreement with the values array and 
uniqueness of label strings.

### DIFF
--- a/maestrowf/datastructures/core/parameters.py
+++ b/maestrowf/datastructures/core/parameters.py
@@ -299,7 +299,10 @@ class ParameterGenerator(SimObject):
             combo = Combination()
             for key in self.parameters.keys():
                 pvalue = self.parameters[key][i]
-                tlabel = self.labels[key].replace(self.label_token,
+                try:
+                    tlabel = self.labels[key][i]
+                except:
+                    tlabel = self.labels[key].replace(self.label_token,
                                                   str(pvalue))
                 name = self.names[key]
                 combo.add(key, name, pvalue, tlabel)
--- a/maestrowf/datastructures/core/parameters.py
+++ b/maestrowf/datastructures/core/parameters.py
@@ -300,10 +300,10 @@ class ParameterGenerator(SimObject):
             for key in self.parameters.keys():
                 pvalue = self.parameters[key][i]
                 try:
-                    tlabel = self.labels[key][i]
-                except:
                     tlabel = self.labels[key].replace(self.label_token,
                                                   str(pvalue))
+                except AttributeError:
+                    tlabel = self.labels[key][i]
                 name = self.names[key]
                 combo.add(key, name, pvalue, tlabel)
             yield combo
--- a/maestrowf/datastructures/yamlspecification.py
+++ b/maestrowf/datastructures/yamlspecification.py
@@ -321,16 +321,18 @@ class YAMLSpecification(Specification):
         Verify that (if globals exist) they conform to the following:
         Each parameter must have:
             1. values
-            2. label
-        Conditions that must be satisfied for a collection fo globals:
+            2. label(s)
+        Conditions that must be satisfied for a collection of globals:
             1. All global names must be unique.
             2. Each list of values must be the same length.
+            3. If the label is a list, its length must match 
+               the value length
         """
         try:
             if self.globals:
                 req_global = set(["values", "label"])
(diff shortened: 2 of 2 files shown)


koning:
@FrankD412 Sorry about the other PR disappearing, I changed the branch and that killed the previous PR. I changed the try/except block and added checks in the yamlspecification.

koning:
Use case:
global.parameters:
  PYIN:
    label:
    - PYIN_0
    - PYIN_1
    - PYIN_2
    values:
    - (S\'res\'\np0\nS\'Low\'\np1\ntp2\n.
    - (S\'res\'\np0\nS\'Med\'\np1\ntp2\n.
    - (S\'res\'\np0\nS\'High\'\np1\ntp2\n.

koning:
@FrankD412  I might have missed the logic on the value length check, was this to check for value, label keys or the length of the values array?

maestrowf/datastructures/core/parameters.py
```diff
@@ -299,8 +299,11 @@ def get_combinations(self):
             combo = Combination()
             for key in self.parameters.keys():
                 pvalue = self.parameters[key][i]
-                tlabel = self.labels[key].replace(self.label_token,
-                                                  str(pvalue))
+                try:
```
FrankD412: I still don't like this `try/catch` method because we could just as easily check to see if something is a list. My suggestion is to use an `if/else` with the `isinstance(x, type)` in this case.

FrankD412:
@koning -- No worries about the disappearing PR, I was just slightly confused. haha

For the length logic, it seems like you got it. The point of that segment is that Maestro can only deal with full lists of user defined combinations. If you're specifying a custom label in the form of a list, you must specify a label for each value the parameter can take (since at this point we cannot be smart about how we generate the label since the user has requested us not to be) -- if that makes sense.

That means that if a list of labels is specified, it must match the length of the list of values (because we're not being naive).

maestrowf/datastructures/yamlspecification.py
```diff
@@ -344,16 +346,30 @@ def verify_parameters(self):
...
                     # If length not set, set it and continue
-                    if value_len == -1:
-                        value_len = len(value)
+                    if values_len == -1:
+                        values_len = len(values)
                         continue
 
                     # Check length. Exception if doesn't match.
-                    if len(value) != value_len:
+                    if len(values) != values_len:
```
FrankD412: Oh, you did miss a conditional here. You'll need to do the following:
```
if isinstance(value["label"], list):
    if len(value) != value_len:
        raise ValueError("Label list for global parameter '{}' is" 
                                   "not the as the list of its values."
                                    .format(name))
```

FrankD412: Interestingly, I think there may be a bug here also. I'm looking into it currently.

maestrowf/datastructures/core/parameters.py
koning: @FrankD412 I was trying to avoid the py2/3 basestr,str issue. I could check for a list and assume it's a string otherwise.

maestrowf/datastructures/yamlspecification.py
koning: @FrankD412 I think that is included, if the values and labels lists are the same length and all of the global values lengths agree, then the the configuration is sound.

FrankD412: @koning -- Oh sorry, I may have been looking at an older version in my sea of tabs... I see it now.

I did some verifying on my end, and your having me review this segment of my original code did expose that I do in fact have a bug here. It looks like you caught it and fixed it. Let me review it one more time to be safe.

maestrowf/datastructures/core/parameters.py
FrankD412: @koning -- I've been assuming `str` which puts us in the realm of ASCII. For now, assume we'll be seeing `str`. It'll be a separate project to go catch `unicode` later.

maestrowf/datastructures/yamlspecification.py
FrankD412: Looks like you fixed it -- thanks for correcting it!

PR #86, opened 2018-04-06
Feature/restart attempts

FrankD412:
Previously we were not exposing the ability for the user to specify the number of successive step restarts (always defaulting to 1). This PR exposes the ability to specify restart limits on the command line.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -120,7 +120,18 @@ def run_study(args):
         LOGGER.error(_msg)
         raise ArgumentError(_msg)
 
-    study.setup(throttle=args.throttle, submission_attempts=args.attempts)
+    # Check if the restart limit is zero or greater:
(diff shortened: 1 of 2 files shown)


maestrowf/maestro.py
```diff
@@ -183,13 +183,18 @@ def setup_argparser():
                                 help="Launch a study based on a specification")
     run.add_argument("-a", "--attempts", type=int, default=1,
                      help="Maximum number of submission attempts before a "
-                     "step is marked as failed.")
+                     "step is marked as failed. [Default: %(default)d]")
+    run.add_argument("-r", "--rlimit", type=int, default=1,
```
dinatale2: Question: Where is this actually passed along to the conductor?

FrankD412: Good catch -- I never pushed in the change. Corrected.

PR #97, opened 2018-05-08
Correction of how dependency ordering happens during study construction.

FrankD412:
Previously Maestro used a spanning tree to find all dependent jobs; however, that method suffered from the issue that steps could be expanded before all their dependencies are. This PR makes it so that Maestro topologically sorts all steps before expanding the Study's DAG into an ExecutionGraph.

### DIFF
--- a/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
+++ b/maestrowf/abstracts/interfaces/schedulerscriptadapter.py
@@ -268,9 +268,9 @@ class SchedulerScriptAdapter(ScriptAdapter):
 
         # If the user is requesting nodes, we need to request the nodes and
         # set up the command with scheduling.
-        _procs = step.run.get("procs")
-        _nodes = step.run.get("nodes")
-        if _procs or _nodes:
+        _nodes = step.run.get("nodes", 0)
+        _procs = step.run.get("procs", 0)
+        if _nodes or _procs:
             to_be_scheduled = True
             cmd = self._substitute_parallel_command(
                 step.run["cmd"],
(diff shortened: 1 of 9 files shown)


FrankD412:
Some notes to keep myself sane... here is some terminology to throw down at this problem; my learnings trying to figure out how to implement this in a few different ways and hitting walls each time. The types of subgraphs/steps we can see in a workflow that I can tell (so far).

- `Linear` - A subgraph that does not use any parameters and results in a DAG that does not need to be expanded. The same can be said for steps within a workflow whose subgraph produces a workflow without parameters.
- `Parameterized` - A node that makes use of parameters or some parameterized component (such as a workspace).
- `Parameter Dependent` - A node or subgraph within a workflow whose expansion depends on an ancestor's parameterized expansion. These are nodes, that may or may not use parameters themselves, are nodes that are required to be expanded because they are directly dependent on other steps who have been expanded. These nodes have a 1:1 mapping with their previous ancestors (so far as I can tell). 
- `Parameter Independent` - A node that, irrespective of being part of a subtree (parameterized or otherwise), does not require expansion. These nodes represent steps that are hubs where parameterized steps get funneled in and technically don't have parent parameters.
- `Parameterized and Parameter Independent` - These nodes are dependent on parameters (or parameterized ancestors) and have a component which funnels a set of parameterized steps. These nodes should be parameterized based on the parameters of the ancestors and itself but not on the parameter independent funnel dependencies.

FrankD412:
To continue my note keeping...

This PR is also going to include a minor rework of where we look for and make use of the `$(OUTPUT_PATH)` keyword `Variable`. The rework of the `Study` construction made it clear that the output path can be substituted much later than it currently is (and how it is currently is kind of suboptimal too).

I'll move the substitution in the StepRecord, since ultimately the output path maps to the workspace for a step so it only makes sense that it should be substituted when the StepRecord gets called. It's making more and more sense that just in time substituting and shifting of output happens in the StepRecord, so I expect more of a shift as I continue refactoring and bug fixing.

PR #112, opened 2018-06-15
Correct a bad import and minor bugs with cancel_jobs

FrankD412:
This PR fixes the `kvs` import so that the module pointed to by `self.kvs` is now the correct one (previously the adapter threw an unknown attribute error when calling the `kvs.get` method). The other fix included is the `.get('status', None)` call in `check_status`.


FrankD412:
Merging, testing on a live project seems to indicate that these fixes are currently working. New PRs will be opened up for other outstanding issues.

PR #113, opened 2018-06-26
Updated the getting started documentation

jsemler:
Updated the getting started documentation to reflect the most recent features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,15 +178,21 @@ The example performs the following workflow locally:
 - Substitute all necessary variables with their serial compilers and make LULESH.
 - Execute a small parameter sweep of varying size and iterations (a simple sensitivity study)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,22 +180,30 @@ The example performs the following workflow locally:
 
 In order to execute the sample study simply execute from the root directory of the repository:
 
-    $ maestro ./samples/lulesh/lulesh_sample1.yaml
+    $ maestro run ./samples/lulesh/lulesh_sample1.yaml
```
FrankD412: You might want to note the macosx version here just in case. `sed` behaves differently between Unix and Mac systems, so they had to be separate specs.

jsemler: Oh, I think is misunderstood. I think you are asking for the macosx spec to also be documented here. The sample spec names have changed. I'll update that.

PR #117, opened 2018-07-06
Workflow setup fix

jsemler:
Fixed a workflow setup issue caused by spaces in the spec name.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -136,7 +136,7 @@ def run_study(args):
     environment.add(Variable("OUTPUT_PATH", output_path))
 
     # Now that we know outpath, set up logging.
-    setup_logging(args, output_path, spec.name)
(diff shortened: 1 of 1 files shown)


maestrowf/maestro.py
```diff
@@ -136,7 +136,7 @@ def run_study(args):
     environment.add(Variable("OUTPUT_PATH", output_path))
 
     # Now that we know outpath, set up logging.
-    setup_logging(args, output_path, spec.name)
+    setup_logging(args, output_path, spec.name.replace(" ", "_"))
```
FrankD412: Just a thought that came to mind -- `.lower()` on the name might also be useful.

jsemler: Sure, I can add that.

jsemler:
PR updated to address reviewer feedback.

Issue #124, opened 2018-08-01
BaseException.message has been deprecated as of Python 2.6


jsemler:
`BaseException.message` has been deprecated. Calls such as `e.message` will result in the following:

- Python 2.7 returns the warning: `DeprecationWarning: BaseException.message has been deprecated as of Python 2.6`.
- Python 3 throws an error: `AttributeError: 'ValueError' object has no attribute 'message'`.

jsemler:
This can be resolved by updating `e.message` to `e.args` or by just referencing the exception instance (e.g.,`logger.exception(e)`).
https://docs.python.org/3/tutorial/errors.html?highlight=exceptions#handling-exceptions

PR #125, opened 2018-08-01
Updated exception messages

jsemler:
This pull request addresses ticket #124: BaseException.message has been deprecated as of Python 2.6.

### DIFF
--- a/maestrowf/conductor.py
+++ b/maestrowf/conductor.py
@@ -199,7 +199,7 @@ def main():
         # Explicitly return a 0 status.
         sys.exit(0)
     except Exception as e:
-        logger.error(e.message, exc_info=True)
+        logger.error(e.args, exc_info=True)
         raise e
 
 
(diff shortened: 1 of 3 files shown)


jsemler:
@FrankD412 -- Sure, let me know what you decide and I can update the PR.

The `Exception.args` does return a tuple. Though when I test the Exception by trying to run a spec file that doesn't exist I get the following message format: `FileNotFoundError: [Errno 2] No such file or directory: 'lulesh_sample1_macosx_test.yaml'`.

If I switch the exception logic to `logger.exception(e)` I get the same message format: `FileNotFoundError: [Errno 2] No such file or directory: 'lulesh_sample1_macosx_test.yaml'`.

FrankD412:
@jsemler -- Huh, not what I expected. It personally doesn't matter to me. I just wanted to verify that the output was correct, ultimately that's what matters.

Issue #107, opened 2018-06-01
Add return codes to the conductor


FrankD412:
The conductor currently terminates normally in all cases that it doesn't throw an exception. While this is useful for knowing that the conductor didn't error out, it's useless for using other scripts to know how a study terminated. This issue is intending to standardize a set of return codes that represent not only the normal termination of the conductor but to also assign codes for the termination of the study.

kcathey:
Is this related to Pre & Post Steps (#98) and the mention it makes of return codes?

FrankD412:
This issue is different -- the difference is as follows:

- Return codes for the conductor designate how the termination of the study itself occurred. So these are things like "terminated properly, all steps completed", or "terminated normally, but with failed steps", etc.
- Return codes for the pre and post steps dictate how the conductor should handle the step. These are return codes that tell the conductor to do things like "restart the step", "rerun the step from scratch", "mark the step as failed", etc. Those get assessed during status check time for any given step.

Hopefully, that clarifies the difference, @kcathey -- if not, keep throwing questions my way!. :-)

kcathey:
@FrankD412 that clarifies it. Thanks!

FrankD412:
@kcathey -- I've assigned this to myself because you're not listed as collaborator. I think I can add you as a Read only user since you're using forks.

kcathey:
Completed with Pull Request #128

PR #128, opened 2018-08-03
#107 Enhance/return codes

kcathey:
These are the changes we discussed.

### DIFF
--- a/maestrowf/abstracts/enums/__init__.py
+++ b/maestrowf/abstracts/enums/__init__.py
@@ -30,7 +30,7 @@
 """Package for providing enumerations for interfaces"""
 from enum import Enum
 
-__all__ = ("JobStatusCode", "State", "SubmissionCode")
+__all__ = ("JobStatusCode", "State", "SubmissionCode", "StudyStatus")
 
 
 class SubmissionCode(Enum):
@@ -65,3 +65,11 @@ class State(Enum):
     TIMEDOUT = 10
     UNKNOWN = 11
     CANCELLED = 12
+
+
+class StudyStatus(Enum):
+    """Workflow status enumeration"""
+    FINISHED = 0
+    RUNNING = 1
+    FAILURE = 2
+    CANCELLED = 3
--- a/maestrowf/abstracts/enums/__init__.py
+++ b/maestrowf/abstracts/enums/__init__.py
@@ -69,7 +69,7 @@ class State(Enum):
 
 class StudyStatus(Enum):
     """Workflow status enumeration"""
-    FINISHED = 0
-    RUNNING = 1
-    FAILURE = 2
-    CANCELLED = 3
+    FINISHED = 0   # The Study has finished successfully, all steps ran
+    RUNNING = 1    # The Study is currently running
+    FAILURE = 2    # The Study has finished, but 1 or more steps failed
+    CANCELLED = 3  # The Study has finished, but was cancelled
--- a/maestrowf/conductor.py
+++ b/maestrowf/conductor.py
@@ -37,6 +37,7 @@ import os
 import sys
 from time import sleep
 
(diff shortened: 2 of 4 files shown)


maestrowf/conductor.py
```diff
@@ -131,8 +132,8 @@ def monitor_study(dag, pickle_path, cancel_lock_path, sleep_time):
                  "sleep time  = %s",
                  pickle_path, cancel_lock_path, sleep_time)
 
-    study_complete = False
-    while not study_complete:
+    study_complete = StudyStatus.RUNNING
```
FrankD412: I recommend renaming this variable just to be clearer.

maestrowf/conductor.py
```diff
@@ -190,14 +194,14 @@ def main():
 
         cancel_lock_path = os.path.join(args.directory, ".cancel.lock")
         logger.info("Starting to monitor '%s'", dag.name)
-        monitor_study(dag, study_pkl[0], cancel_lock_path, args.sleeptime)
+        study_status = monitor_study(dag, study_pkl[0], cancel_lock_path, args.sleeptime)
```
FrankD412: This line fails flake8 checks -- it's longer than 79 characters.

FrankD412:
@kcathey -- I'd also recommend rebasing your branch and resolving conflicts that may come up.

maestrowf/maestro.py
```diff
@@ -208,7 +208,9 @@ def run_study(args):
             # Launch in the foreground.
             LOGGER.info("Running Maestro Conductor in the foreground.")
             cancel_path = os.path.join(path, ".cancel.lock")
-            monitor_study(exec_dag, pkl_path, cancel_path, args.sleeptime)
+            # capture the StudyStatus enum to return
+            study_status = monitor_study(exec_dag, pkl_path, cancel_path, args.sleeptime)
```
FrankD412: I believe this one is also longer than 79 characters.

maestrowf/conductor.py
kcathey: Yup you're right

maestrowf/conductor.py
kcathey: I'll take a look

maestrowf/maestro.py
kcathey: I'll take a look

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -645,7 +645,7 @@ def execute_ready_steps(self):
             # Just return for now, but we'll need a way to signal that there
             # are no more things to run.
             logging.info("'%s' is complete. Returning.", self.name)
-            return True
+            return StudyStatus.FINISHED
```
FrankD412: @kcathey and I were talking about this condition. I found that the conductor returned 0 (FINISHED) status even when cancelled. @kcathey pointed out that this conditional gets triggered without checking for cancelled state. Kevin is going to look into a refactor to clean this up because I had initially done this to assess the state entering and leaving the status check to avoid needlessly waiting for another round of updates.

maestrowf/abstracts/enums/__init__.py
```diff
@@ -65,3 +65,11 @@ class State(Enum):
     TIMEDOUT = 10
     UNKNOWN = 11
     CANCELLED = 12
+
+
+class StudyStatus(Enum):
+    """Workflow status enumeration"""
+    FINISHED = 0
```
FrankD412: For documentation purposes, it might be good to add inline comments here for the definitions of the enums. I plan to go back and update the other ones so their definitions are clear.

kcathey: So something like this: FINISHED = 0   # The Study has finished successfully, all steps ran

PR #129, opened 2018-08-06
Addition of custom parameter generation when running studies.

FrankD412:
This PR adds the ability for a user to use custom code for generating parameters. In all cases, metadata about parameters is now dumped. In order to use this feature, a user simply uses the command line ```maestro run --pgen <path to .py> <path to study YAML>```. Within the file specified by the ```--pgen``` flag, the function `get_custom_generator()` must be defined and must return a `ParameterGenerator` instance.

### DIFF
--- a/maestrowf/datastructures/core/parameters.py
+++ b/maestrowf/datastructures/core/parameters.py
@@ -347,3 +347,17 @@ class ParameterGenerator(SimObject):
         params = set()
         self._get_used_parameters(step.__dict__, params)
         return params
+
+    def get_metadata(self):
+        """
+        Produce metadata for the parameters in a generator instance.
+
+        :returns: A dictionary containing metadata about the instance.
+        """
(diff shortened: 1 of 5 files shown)


FrankD412:
@jsemler -- Would you mind looking at the metadata generated? I went with what I thought was reasonable, but if you could provide any insight into what might be useful please let me know.

FrankD412:
That's actually a good thought -- I hadn't thought to document the path. Actually, I was also considering copying the code file itself into the directory but wasn't sure if that was overkill.

jsemler:
Having the path to the `-pgen` file in the metadata would be useful. I think copying the code file could potentially be helpful for both archiving and reproducibility. I will have to think on that more.

FrankD412:
@jsemler -- I've made the requested changes, can you give it another pass?

PR #131, opened 2018-08-08
Bugfix that would cause the ExecutionGraph to not update when cancelled.

FrankD412:


FrankD412:
@kcathey -- This bug was introduced because it looked like you had added a short circuit to `execute_ready_steps`  which prevented the states in the DAG from being updated. I simply removed the short circuit so that it can update the statuses. The check for running state in `monitor_study` prevents the conductor itself from sticking around longer than it should.

PR #132, opened 2018-08-08
Addition of a LULESH study specification with heavier comments.

FrankD412:


FrankD412:
@jsemler @kcathey @tadesautels  -- feel free to comment as I commit. I'm trying to make comments as clear as possible.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,199 @@
...
+            sed -i 's/^#CXXFLAGS = -g -O3 -I/CXXFLAGS = -g -O3 -I/' ./Makefile
+            make clean
+            make
+          # depends: A list of task names that this step is dependent on.
+          # The 'make-lulesh' step is not dependent on other tasks, and so will
+          # be executed at the start of the study.
+          depends: []
+
+      # 'run-lulesh' is an example of a parameterized step because the task
+      # contains parameters (more on those later).
```
jsemler: Consider clarifying that the parameters being referenced in the step are `$(SIZE)` and `$(ITERATIONS)`.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,199 @@
...
+# section is meant primarily for documentation purposes so that when a
+# spcecifcation is passed to other users they can gleam a general understanding
+# of what this study is meant to achieve.
+#-------------------------------
+# Required keys:
+#   name - Name of the study
+#   description - Description of what this study does.
+#-------------------------------
+# NOTE: You can add other keys to this block for custom documentation. Maestro
+# currently on looks for the required set.
```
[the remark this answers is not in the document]
FrankD412: Good catch. Fixed.

samples/documentation/lulesh_sample1_unix_commented.yaml
FrankD412: Good point -- added.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -187,10 +191,41 @@ study:
       # on 'run-lulesh_*' which declares dependency on ALL combinations of
       # 'run-lulesh').
 
+####################################
+# Parameter Block
+####################################
+# The parameter block contains all the things we'd like to vary in the study.
+# Currently, there are two modes of operating in the specification:
+# 1. If a parameter block is specified, the study is expanded and considered a
```
jsemler: Does the parameter also have to be referenced in a step?

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -187,10 +191,41 @@ study:
...
+#   parameterized study.
+# 2. If a parameter block is not specified, the study is treated as linear and
+#    the resulting study is not expanded.
+#
+# There are three keys per parameter:
+# 1. A list of values that the parameter takes.
+# 2. A label that represents a "pretty printed" version of the parameter. The
+#    parameter values is specfied by the '%%' moniker (for example, for SIZE --
+#    when SIZE is equal to 10, the label will be 'SIZE.10').
+# 3. (Optional) A list of custom names for each value in the values list.
```
jsemler: If `names` is specified is that used instead of the parameter `values` when creating the workspace directories?

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
+####################################
+# Description Block (Required)
+####################################
+# The description block is where the description of the study is placed. This
+# section is meant primarily for documentation purposes so that when a
+# spcecifcation is passed to other users they can gleam a general understanding
```
tadesautels: "gleam" --> "glean"

tadesautels: Also, explicit mention that this block is required.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
...
+####################################
+env:
+    #-------------------------------
+    # Variables
+    #-------------------------------
+    # Static values that the workflow substitutes into steps and are similar in
+    # concept to Unix environment variables. These variables are not dependent
+    # on values in the environment and so are more portable.
+
+    # NOTE: These values are substituted as static strings meaning that other
```
tadesautels: Drop first "other"

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
...
+    # variables should not used in other variables.
+    #-------------------------------
+    variables:
+        # OUTPUT_PATH is a keyword variable that Maestro looks for in order to
+        # set a custom output path for the study workspace. If not specified,
+        # OUTPUT_PATH is assumed to be the path where Maestro was launched from.
+        OUTPUT_PATH: ./sample_output/lulesh
```
tadesautels: I have found this one a bit confusing; it's really more of a workspace_root, rather than an output_path. 
Renaming this variable would probably be best, but the description should certainly say that this is where ALL of maestro's temporary files will go, in addition to any output, and typically within a custom-named, date-stamped directory for this run of maestro.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
...
+            make clean
+            make
+          # depends: A list of task names that this step is dependent on.
+          # The 'make-lulesh' step is not dependent on other tasks, and so will
+          # be executed at the start of the study.
+          depends: []
+          # NOTE: There are additional step entries for setting up a step for
+          # remote execution. For examples see 'launcher_tokens.yaml' in the
+          # 'samples/documentation' subdirectory in the maestrowf repo.
+
```
tadesautels: I presume you're going to eventually have a more complete, formal list of all of the options, right?

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
...
+      # NOTE: The 'depends' entry in a specification does not use parameterized
+      # names -- instead Maestro attempts to infer the expansion of the graph
+      # based on context via looking to see if a step uses parameters, workspaces,
+      # and if the step depends on other steps that are parameterized.
+      # NOTE: Second note -- this example lacks a type of step. It is possible
+      # to have a step that is expanded because its dependencies have been
+      # expanded. For example, if you wanted to run individual post processing
+      # after each 'run-lulesh' step, you might have a step called
+      # 'post-process-run' that is dependent on 'run-lulesh'. This notation
```
tadesautels: This should explicitly say, "the job step specification would then contain the value 'depends : [run_lulesh]'."

If this is what you mean.

samples/documentation/lulesh_sample1_unix_commented.yaml
```diff
@@ -0,0 +1,234 @@
...
+      # run: Properties that describe the actual specification of the task.
+      run:
+          # cmd: The actual task to be executed. (Required)
+          cmd: |
+            cd $(LULESH)
+            sed -i 's/^CXX = $(MPICXX)/CXX = $(SERCXX)/' ./Makefile
+            sed -i 's/^CXXFLAGS = -g -O3 -fopenmp/#CXXFLAGS = -g -O3 -fopenmp/' ./Makefile
+            sed -i 's/^#LDFLAGS = -g -O3/LDFLAGS = -g -O3/' ./Makefile
+            sed -i 's/^LDFLAGS = -g -O3 -fopenmp/#LDFLAGS = -g -O3 -fopenmp/' ./Makefile
+            sed -i 's/^#CXXFLAGS = -g -O3 -I/CXXFLAGS = -g -O3 -I/' ./Makefile
```
kcathey: I know that this out of our control, but these flags should really be configured using a configure script that lulesh supplies. We might want to send that back to the lulesh folks. Maestro would then not need to know the defaults of how to build lulesh, just run configure make clean and make.

FrankD412: That's a fine point. I agree. At the very least maybe a `make serial` or `make parallel`.

samples/documentation/lulesh_sample1_unix_commented.yaml
FrankD412: Yeah, that the eventual goal. The comments here are meant to help users make sense of this particular spec and to get the generals in place.

samples/documentation/lulesh_sample1_unix_commented.yaml
FrankD412: This comment prompted me to go back and realize I was not remembering the functionality correctly. This'll be fixed in the next update.

samples/documentation/lulesh_sample1_unix_commented.yaml
FrankD412: I added a statement to this that a parameter does not necessarily need to be used.

samples/documentation/lulesh_sample1_unix_commented.yaml
FrankD412: I corrected this in my last commit -- let me know if it's still unclear.

samples/documentation/lulesh_sample1_unix_commented.yaml
tadesautels: Is the question if global.parameters implies the non-Linear expansion, even if the parameters are unused?

PR #133, opened 2018-08-15
Correction to a bug for attempting a restart when a restart isn't specified.

FrankD412:
Previously, the conductor made the assumption that if a step timed out that the user specified a way to restart. That assumption does not necessarily hold because a restart can not be specified. This PR corrects that assumption by adding the capability to check if a record can be restarted and does the appropriate book keeping if it cannot be.

### DIFF
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -104,6 +104,18 @@ class _StepRecord(object):
 
         return retcode
 
+    @property
+    def can_restart(self):
+        """
+        Get whether or not the record can be restarted.
+
+        :returns: True if the record has a restart command assigned to it.
+        """
+        if self.restart_script:
+            return True
+
+        return False
+
     def _execute(self, adapter, script):
         if self.to_be_scheduled:
             retcode, jobid = adapter.submit(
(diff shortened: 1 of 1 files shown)


FrankD412:
@jsemler -- You previously had steps that timed out, was the timeout properly reflected in the status?

jsemler:
Yes, the timeouts are reflected. Looking through the status again I'm noticing some of the steps don’t include the restart count in the status.csv.

FrankD412:
@jsemler -- Can you provide an example?

jsemler:
It looks like the restart count is in the status.csv file, but there is an issue with it displaying via the console for some reason:
`run-sedov   run-sedov     RUNNING      -1 day          22:29:02.592852  0:01:00.479318              2018-08-16 19:03:40.142983  2018-08-16 17:31:42.256517  2018-08-16 17:32:42.735835`

FrankD412:
Yeah, I was just looking at the log. The weird thing is the conductor launches it, then for a while reports it's complete and that it's being skipped. Then after some point, it's picked up again. Still not sure why.

FrankD412:
@jsemler -- I'm thinking the bug fix for the `ready_steps` bug is a new PR. If you can, review and comment on this PR in terms of fixing the original issue? I'll start a new branch for the other bug.

PR #136, opened 2018-08-17
Throttled workflows push steps into the ready queue multiple times.

FrankD412:
Originally from https://github.com/LLNL/maestrowf/pull/133#issuecomment-413932611:

@jsemler -- Alright, after a good bit of digging around, I think I've narrowed the bug down. I apologize if this explanation is somewhat convoluted, but I'll do what I can to make it clear.

In the `ExecutionGraph` method `execute_ready_steps`, there are a number of back end sets that track various aspects of execution, including failed steps, in progress steps, completed steps, dependency completions, and various other things (which for this discussion aren't important). The major thing that gets tracked are the ready steps which are in a `deque` object. When a step is deemed to be ready, its `StepRecord` object is pushed to the `ready_steps` queue. However, in the `execute_ready_steps` method, the following code currently exists (line 785 in executiongraph.py):

```
# If the gating dependencies set is empty, we can execute.
if not self._dependencies[record.name]:
if key not in self.ready_steps:
    logger.debug("All dependencies completed. Staging.")
    self.ready_steps.append(record)
else:
    logger.debug("Already staged. Passing.")
    continue
```

I noticed that in the log you provided that steps that were repeated would always trigger the "All dependencies completed. Staging." log message and I never saw the passing message in the `else`. That's likely because when I refactored and cleaned up the method last time, I made the transition to tracking records by their name instead of the object in order to clean up a lot of the `record.name` referencing I was doing. Somewhere along the way, it looks like I forgot to transition the `ready_steps` queue to the same structure and so it goes to check if the key is in the `deque` and never finds it since it's a queue of `StepRecord` objects.

I don't expect the fix to this will be difficult, so I'll try and have it completed today. We can still go through it if you'd like.

Edit: As an extra note, I think this issue only appears when throttling because that is the only time you'll have steps queued that are still in the `INITIALIZED` state. If not throttling, you'd never run into a duplicate in the queue since those steps would be in at least `PENDING` when submitted.

### DIFF
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -828,7 +828,13 @@ class ExecutionGraph(DAG):
         else:
             # Compute the number of available slots we have for execution.
             _available = self._submission_throttle - len(self.in_progress)
+            # Available slots should never be negative, but on the off chance
+            # we are in a slot deficit, then we will just say none are free.
             _available = max(0, _available)
+            # Now, we need to take the min of the length of the queue and the
+            # computed number of slots. We could have free slots, but have less
+            # in the queue.
+            _available = min(_available, len(self.ready_steps))
             logger.info("Found %d available slots...", _available)
 
         for i in range(0, _available):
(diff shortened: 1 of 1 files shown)


jsemler:
@FrankD412 -- I ran a few tests with and without the `--throttle` flag and the `-1 day` bug did not occur. Is there anything else you’d like me to test?

FrankD412:
@jsemler -- It's good that didn't appear since it was the main symptom. If your tests indicate that a study runs to completion with and without throttling then this issue should be fixed. It appeared from the logs last week that steps were only run once now, so we should be good.

PR #144, opened 2018-09-16
Cleans up GitDependency errors a but to not rely on return codes.

### LINKED ISSUE #140: Git dependency error message when no internet access
When Git is unable to reach the internet, Maestro returns the same error message as if the destination directory were already present; it appears that Git returns the same error code in both cases.

See gitdependency.py, lines 151 - 162, and the following StackOverflow thread:
https://stackoverflow.com/questions/4917871/does-git-return-specific-return-error-codes

One possible check would be to try pinging the URL: a successful ping returns status 0 when stopped, whereas a ping without an internet connection returns code 68.

FrankD412:
Fixes #140 -- addition of the `ping_url` utility function and some minor clean up of the `GitDependency` error checking in `acquire`.


maestrowf/datastructures/environment/gitdependency.py
```diff
@@ -146,21 +147,30 @@ def acquire(self, substitutions=None):
                 self.path = substitution.substitute(self.path)
                 self.url = substitution.substitute(self.url)
 
+        # Check that we are connected to the internet.
+        logger.debug("Pinging Google to verify internet connectivity...")
+        ping_url("https://www.google.com")
```
jsemler: Does a failed connection automatically cancel the workflow?

FrankD412: Yeah, if `ping_url` can't find the URL it throws an exception. We acquire before starting the workflow, so it'd not even get to launching the conductor.

maestrowf/utils.py
```diff
@@ -153,3 +155,24 @@ def make_safe_path(*args):
...
+    """
+    try:
+        response = urlopen(url)
+    except HTTPError as e:
+        LOGGER.error("Error fulfilling HTTP request. (%s)", e.code)
+        raise e
+    except URLError as e:
+        LOGGER.error(
+            "Check specified URL (%s) and that you are connected to the "
+            "internet. (%s)", url, e.code)
```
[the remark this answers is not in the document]
FrankD412: Oh, I saw that somewhere and thought it existed. I'll go ahead and look into this.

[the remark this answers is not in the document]
FrankD412: I've reverted to just throwing the exception.

FrankD412:
I've implemented a new way of testing that @jsemler pointed out. I now use the the `git remote-ls <url>` command. If the command is able to connect (I ignore its output) it returns a code of 0. Otherwise, it simply returns 128. I used that check to get away from having to ping Google.

As a side, I did leave the utility function in utils because it can be useful in the future for other checks or dependencies.

FrankD412:
@tadesautels -- It looks like you're talking about two different messages, but I think that's probably a good idea.

FrankD412:
@tadesautels -- I went through and removed the error codes.

maestrowf/datastructures/environment/gitdependency.py
FrankD412: I've changed this to check the repo url directly.

PR #145, opened 2018-09-21
Addition of user enabled workspace hashing

FrankD412:
This PR introduces hashing of the combination substrings for steps that use parameters in parameterized studies. I've settled for md5 since it produces the shortest strings.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -216,7 +216,7 @@ def run_study(args):
     study.setup_environment()
     study.configure_study(
         throttle=args.throttle, submission_attempts=args.attempts,
-        restart_limit=args.rlimit, use_tmp=args.usetmp)
+        restart_limit=args.rlimit, use_tmp=args.usetmp, hash_ws=args.hashws)
 
     # Stage the study.
     path, exec_dag = study.stage()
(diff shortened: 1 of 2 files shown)


FrankD412:
@jsemler -- Try this out. The meta data and everything is updated accordingly, so it shouldn't change anything in terms of functionality around it for parsing metadata. Simply add `--hashws` to the `maestro run` command.

jsemler:
Some of the post processing steps in the LULESH example are failing when `--hashws` is specified. It looks like the label attribute doesn’t reference the hashed workspace directory name.

FrankD412:
It sounds like you're talking about the workspace tokens? Do you have an error that appeared?

FrankD412:
Got it -- I'm inclined to think that this is because the sample specification was written without the hashing in mind which causes the `grep` command for the label to fail. It leads me to think that this is a subtlety of using the hashing functionality, but since it's optional it's something the user should account for.

maestrowf/maestro.py
```diff
@@ -319,6 +319,9 @@ def setup_argparser():
     run.add_argument("-fg", action="store_true", default=False,
                      help="Runs the backend conductor in the foreground "
                      "instead of using nohup. [Default: %(default)s]")
+    run.add_argument("--hashws", action="store_true", default=False,
+                     help="Enable hashing of subdirectories in parameterized "
```
jsemler: Consider adding a note about the use of parameter label attributes in a study when `--hashws` is specified.

PR #150, opened 2018-10-16
Rework the LocalScriptAdapter to more closely mimic an actual scheduler.

FrankD412:
NOTE: This PR is turning into a somewhat extensive change, so I wanted to throw up a PR to better track and discuss the changes that are happening and maybe see if anyone had any thoughts.

This PR captures all related changes to the `LocalScriptAdapter` that are aiming to have it mimic an actual scheduler. The current plan is to implement a threaded `ExecutionPool` that actively updates and stores records of locally executing processes kicked off through the conductor. The conductor can then pass a list of process identifiers to retrieve the status just like it would for the SLURM adapter or any other scheduler. This requires a `Singleton` instance of the `ExecutionPool` which provides the backend lifting for the `LocalScriptAdapter`.  So far, this has necessitated the use of a ReadWrite lock that would lock the backend status structures (again, this is something that any scheduler does automatically).

### DIFF
--- a/maestrowf/datastructures/utilities/executionpool.py
+++ b/maestrowf/datastructures/utilities/executionpool.py
@@ -1,0 +1,38 @@
+"""A module for containing data structures for process pool management"""
+
+import logging
+from Queue import Queue
+from threading import Event, Lock, Thread
+
+from maestrowf.abstracts import Singleton
+from maestrowf.utils import start_process
+
+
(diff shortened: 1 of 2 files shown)


FrankD412:
I've come to the determination that this PR is outside the scope of Maestro. Because of the complicated internals that something like this could take and the fact that such a program could be useful in other circumstances, I've opted to pass that functionality up.

PR #151, opened 2018-10-20
YAML Specification v2.0

FrankD412:
This PR is for discussion and iterative improvements for a version 2.0 of the YAML specification. Some of the goals I'm aiming for with v2.0:
- Clean up some possible misnomers and other verbiage that might be confusing.
- Restructure the representation of a step to be cleaner and allow for more flexible definition of used resources.
- General clean up to make it a little more streamlined.


samples/lulesh/lulesh_flux_spec_v2.0.yaml
```diff
@@ -0,0 +1,70 @@
...
+        outfile: $(SIZE.label).$(ITERATIONS.label).log
+
+    dependencies:
+      git:
+        - name: LULESH
+          path: $(OUTPUT_PATH)
+          url: https://github.com/LLNL/LULESH.git
+
+study:
+    make-lulesh:
```
jsemler: This is an interesting change to the YAML description. It makes it a little less verbose. I do miss the list (`-`) syntax though. It made it easier to make out the individual steps at a glance.

FrankD412: It's something I've been going back and forth on. I figured it'd be good to just throw the kitchen sink into the rework and see what stuck. My main concern is with how easy it is to mess up the style with the `-` list notation, but I can see merits in it either way.

Thoughts? @gonsie @kcathey @dinatale2 @tadesautels

FrankD412: On that note, I was considering something similar for the dependencies as well.

```
dependencies:
    git:
        LULSEH:
            path: $(OUTPUT_PATH)
            url: https://github.com/LLNL/LULESH.git
```

dinatale2: I agree that it might be easy to mess up the list notation if you keep the list notation, but consider that your users are likely used to it. I'd suggest taking some existing specs and converting them to the new format to see if you're missing anything. You can also get feedback from your users once you mock up existing ones.

FrankD412: @kcathey -- The depends entry is already a list, so that's handled.

In terms of default values for the batch settings, I've been considering that. However, I have also gotten feedback that the batch settings at the top of the file feel kind of out of place. Like it should all be in the steps -- I'm not remembering 100% what the user said, @jsemler do you happen to recall?

One thing I keep revisiting is just having users specify general values in the specification per step and then having more platform specific information passed in separately (or calculated using cluster information).

I also had another thought on the dependencies section of the specification -- I really don't like that it's three layers deep by the time you specify name, type, then path and other details. As someone who writes code for a living, I feel like it's just stylistically gross and it kind of disrupts the flow of writing my code (or specification in this case) because it becomes harder to track the scope I'm writing in. I had brought this new format to @jsemler but never wrote it down anywhere, it's as follows:

```
dependencies:
    LULSEH:
        type: git
        path: $(OUTPUT_PATH)
        url: https://github.com/LLNL/LULESH.git
```

The new format accomplishes two things primarily: makes the spec less tab heavy and also makes it so the name of the dependencies become the focus.

samples/lulesh/lulesh_flux_spec_v2.0.yaml
```diff
@@ -10,10 +10,10 @@ env:
         outfile: $(SIZE.label).$(ITERATIONS.label).log
 
     dependencies:
-      git:
-        - name: LULESH
-          path: $(OUTPUT_PATH)
-          url: https://github.com/LLNL/LULESH.git
+      LULESH:
```
FrankD412: This format is just my suggested one from my previous comment. I just wanted to put it here to have it in the spec to see how it looks. @jsemler and @kcathey -- thoughts?

samples/lulesh/lulesh_flux_spec_v2.0.yaml
jsemler: @FrankD412 -- I don't think there was any additional information regarding the batch settings feedback.

I think reducing the levels of indentation for listing dependencies is helpful. One advantage of the current version is it groups dependencies by type. I’m not sure how readable the updated format will be for larger lists of dependencies.

FrankD412: IMO, I think the updated format is more readable because it is explicit -- more so in longer lists of dependencies. If you have a list of items long enough you'd not need to scroll anywhere to know what type it was since it's 1:1 mapped directly in the entry.

jsemler: I meant readable from an organizational standpoint of all the dependencies. Though, that might not matter as much. Having the dependency type listed as part of the dependency is helpful. There is redundancy if multiple dependencies have the same type though (not sure how much that matters).

samples/lulesh/lulesh_flux_spec_v2.0.yaml
jsemler: The indentation looks inconsistent with the rest of the `env` section. Otherwise, it looks good.

FrankD412:
@jsemler @kcathey @gonsie -- I've been trying to think of a clean way to allow users to set actual environment variables either via the spec or some other means. It's been bothering me since a use case came up for a code (I think it was gromacs) that required environment variables to point the binary to some utility binaries or something. I ended up handling it with labels, but it felt a little repetitive. I'm looking for thoughts on if that's a good idea and if it so how to cleanly make it available without confusing the user.

kcathey:
> @jsemler @kcathey @gonsie -- I've been trying to think of a clean way to allow users to set actual environment variables either via the spec or some other means. It's been bothering me since a use case came up for a code (I think it was gromacs) that required environment variables to point the binary to some utility binaries or something. I ended up handling it with labels, but it felt a little repetitive. I'm looking for thoughts on if that's a good idea and if it so how to cleanly make it available without confusing the user.
> 
> If the environment variables route is something worth considering, I was toying with the idea of moving the dependencies section out of env and only leaving variables and labels in there. But then it still suffers from the same problem of how to specify something to be set in the environment (again, just some thoughts I had based on a use case that's been bothering me).

I almost think you need a section like: EXECUTION_ENVIRONMENT in which people can set their environment variables, it would need to be both global and task specific. Then when you generate the script embed any variables into the script before actually calling the user's desired runtime.

FrankD412:
@kcathey -- Might it be worth passing those in as a command line parameter? I've had the idea of having the conductor dump out the environment when it goes to boot up the study. That would be a closer step to reproducibility because it'd allow someone to outright pass in the environment.

I do disagree with the notion that the environment variables should be step specific. That increases the places where a variable might need to be updated and promotes people reusing variable names when they might not want to.

kcathey:
@FrankD412 -- The conductor dumping the environment to the study as it starts is potentially useful, and a good first step. To truly get reproducibility they should be defined somewhere, it could be in the spec, or it could be in a separate .env type file.

I can see a situation where the environment variable name needs to be passed to different steps, but needing to have a different value to get codes to work as desired. I don't think is best practice from the codes perspective, but it could be the world in which we have to play.

FrankD412:
This PR is super stale at this point, and the discussion on this has fizzled out. Closing.

PR #153, opened 2018-10-21
Converts specification look up into a factory.

FrankD412:
Since we're looking to update the YAML specification, it's probably about time we convert to a factory for specifications. The factory let's us retain reverse compatibility to older versions of the YAML specification and support other types of specifications in the future.


maestrowf/datastructures/specifications/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+            "1.0":      (".yamlspecification", "YAMLSpecification"),
+        },
+    }
+
+    @classmethod
+    def get_specification(cls, spec_name, version="latest"):
+        """
+        Look up and retrieve a Specification by name.
+
+        :param spec_name: Name of the Specification to find.
```
jsemler: Is `spec_name` referring to the specification format (e.g., yaml)?

FrankD412: Yes and no -- I couldn't figure out a good name for a Maestro YAML specification, but that could also be something like "cwl" for Common Workflow Language and things like that. I'm definitely welcome to suggestions either for naming or a better way to handle the "naming" to make it clearer.

maestrowf/datastructures/specifications/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+class SpecificationFactory(object):
+    """A factory class to retrieve different types of Specifications."""
+
+    _classes = {
+        "yaml": {
```
dinatale2: Question: This means you're going to support other specifications in the future?

maestrowf/datastructures/specifications/__init__.py
dinatale2: How about `spec_type` or `spec_kind`?

FrankD412: @dinatale2 -- I updated to `spec_type`.

PR #156, opened 2018-10-31
Removal of _stage_linear since it is now not needed.

FrankD412:
@tadesautels and @jsemler both ran into the same bug related to linear study staging. It seems that the previous `_stage_linear` in the `Study` class was outdated since the new `_stage_parameterized` was introduced. It seems like it was entirely an oversight, but it now seems like it's redundant since the parameterized staging is general enough to handle both cases. The simplest solution to this bug seems to just move `_stage_parameterized` to be the general solution and to remove `_stage_linear` entirely.

@tadesautels and @jsemler -- Can you two confirm that this fix works for the studies that triggered your bug?


FrankD412:
#157 corrects a bug that was reproduced while working on this PR. The bug was related to having a space in the study name due to not having all paths passed through the `make_safe_path`. #157 details the fix and enhancements made and should probably be merged before this PR.

PR #160, opened 2018-11-05
Correction to fix the format of output status time to avoid a comma that breaks printing.

### LINKED ISSUE #159: Long run time breaks status file parsing
When the run time exceeds 24hrs, `1 day, ` gets added to the time stamp string, which breaks the status.csv file parsing due to the extra `,`.

FrankD412:
Fixes #159 -- long running jobs can reach the day range and breaks the formatted print because the time difference contains a comma.

### DIFF
--- a/maestrowf/utils.py
+++ b/maestrowf/utils.py
@@ -41,9 +41,24 @@ import time
 LOGGER = logging.getLogger(__name__)
 
 
+def get_duration(time_delta):
+    """
+    Covert durations to HH:MM:SS format.
+
+    :params time_delta: A time difference in datatime format.
+    :returns: A formatted string in HH:MM:SS
+    """
+    duration = time_delta.total_seconds()
+    hours = int(duration / 3600)
+    minutes = int(duration % 3600 / 60)
+    seconds = int((duration % 3600) % 60)
+
+    return "{:02d}:{:02d}:{:02d}".format(hours, minutes, seconds)
+
+
 def generate_filename(path, append_time=True):
     """
-    Utility function for generating a non-conflicting file name.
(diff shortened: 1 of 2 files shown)


FrankD412:
The new format is now something like ```'2 days 08h:13m:14s'```

maestrowf/utils.py
```diff
@@ -41,9 +41,26 @@
...
+    :params time_delta: A time difference in datatime format.
+    :returns: A formatted string in HH:MM:SS
+    """
+    duration = time_delta.total_seconds()
+    days = int(duration / 86400)
+    hours = int((duration % 86400) / 3600)
+    minutes = int((duration % 86400 % 3600) / 60)
+    seconds = int((duration % 86400 % 3600) % 60)
+
+    return "{:d} days {:02d}h:{:02d}m:{:02d}s" \
```
kcathey: I'm actually wondering if the format should be like -  2d:3h:3m:3s that makes it more consistent and easier to parse since everything is 2 digits and a letter.

FrankD412: You know, that’s a good point. I had originally had that and then changed it. But the silly thing is that I don’t necessarily have to limit the days to 2 digits like the rest. I’ll make a new commit that changes the format.

FrankD412: You know, that’s a good point. I had originally had that and then changed it. But the silly thing is that I don’t necessarily have to limit the days to 2 digits like the rest. I’ll make a new commit that changes the format.

FrankD412:
@jsemler and @kcathey I just changed the format. It should work, but could one of you verify?

PR #163, opened 2018-11-14
Added study confirmation message

jsemler:
Added a confirmation message after successfully launching a study.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -282,6 +282,8 @@ def run_study(args):
             LOGGER.debug(" ".join(cmd))
             start_process(" ".join(cmd))
 
+    print("Study launched successfully.")
+
     return 0
 
 


kcathey:
This is useful, but I'm wondering if we want to add a logger that will always print to the console. To keep consistent on how messages are presented. We also might want this message sent to other log levels as well.

FrankD412:
@kcathey -- I'm probably making a project to change the user facing output away from the logging messages. Currently all that's really dumped is the expansion information, and I think more user friendly output is useful. While the logging is useful, it also bloats the terminal output with timestamps and function names.

jsemler:
@kcathey -- I agree. I think that would be helpful.

It sounds like @FrankD412 is planning some rework to the output. Let me know if I should update this PR to use a logger.

Issue #167, opened 2018-12-21
Convert ScriptAdapterFactory to Discoverable Plugin Model


kcathey:
Using the methods described in https://packaging.python.org/guides/creating-and-discovering-plugins/ convert the ScriptAdapterFactory to use those methodologies to register ScriptAdapters.

The idea is to enable users to develop their own ScriptAdapters for various schedulers, or their own proprietary scheduler. #

PR #170, opened 2019-01-18
Script Adapter Plugin (#167)

kcathey:
This pull request is the implementation for Issue #167.

ScriptAdapters have been converted to load dynamically, converting them to a plugin type infrastructure. The maestro.interfaces.script package loads all modules and all classes underneath it, and registers them into the ScriptAdapterFactory with a key.

Pytest was used to guarantee that existing functionality was preserved.


tox.ini
```diff
@@ -4,20 +4,26 @@
...
 
-[testenv:nose]
-deps = nose
-commands =
-	nosetests
+[testenv]
+deps:
+	pytest
+commands:
+	pytest {posargs}
```
FrankD412: I had initially planned to use nose for testing -- to my understanding, people prefer one or the other usually but I recall reading that nose was not being as supported or something to that effect. I'll have to google around to catch up on what is considered standard. I don't personally have a preference either way, but we can sort that out later.

Related, since we now require pytests, we need to include it in the `requirements.txt` so that developers will automatically pull it when they setup initially.

tests/interfaces/script/test_fluxscriptadapter.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+###############################################################################
+from maestrowf.interfaces.script.fluxscriptadapter import FluxScriptAdapter,\
```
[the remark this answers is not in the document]
kcathey: Not a problem.

tox.ini
kcathey: Pytest is added to the requirements.txt.

As for the pytest/nose I prefer pytest, but do not care either way. I did run these tests through nose and they worked there as well so switching should not be a problem.

tests/interfaces/script/test_fluxscriptadapter.py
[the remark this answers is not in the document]
kcathey: @FrankD412 I've done these updates as requested.

tox.ini
FrankD412: Oh -- I was in the wrong environment. There's pytest. We'll stick with pytest for now then, I just removed nose from the requirements.

FrankD412:
Things look good to me -- are there any other changes you're looking to make @kcathey?

kcathey:
At this point in time I think this is pretty much done.

PR #171, opened 2019-01-22
PyYAML vulnerability fix

FrankD412:
There is a known vulnerability in PyYAML version >= 2.4b1 -- this PR adds the fix so that pip and Pipenv install the versions that fulfill the condition to fix the vulnerability.


lucpeterson:
This throws a deprecation warning now. Need to update yaml.load to include the loader

PR #177, opened 2019-03-19
Change to transition adapter returns to Record objects.

FrankD412:
After some discussion with @lucpeterson @jsemler @koning @robinson96 , it was decided that adapters should rely on `Record` classes to improve the ability of adapters to pass back extra information. Their use case is specifically referring to the `submit` method of adapters, but the notion of records is useful to all adapter calls. I've gone ahead and ported all methods to a record class.

### DIFF
--- a/maestrowf/abstracts/containers/__init__.py
+++ b/maestrowf/abstracts/containers/__init__.py
@@ -1,0 +1,50 @@
+###############################################################################
+# Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+# Produced at the Lawrence Livermore National Laboratory
+# Written by Francesco Di Natale, dinatale3@llnl.gov.
+#
+# LLNL-CODE-734340
+# All rights reserved.
+# This file is part of MaestroWF, Version: 1.0.0.
+#
+# For details, see https://github.com/LLNL/maestrowf.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


FrankD412:
I've marked the PR as "In Progress" because I'd like feedback as I develop this. Please take a look and offer feedback as soon as you think of something.

FrankD412:
I've gotten submission done, but I've hit a stumbling block on cancel. Most schedulers allow you to cancel all jobs at once, which means that you're cancelling a bunch of them all at once without a way to check their individual statuses. Most schedulers I'd imagine would just expect cancel to work, but from the outside we have a few options each with varying degrees of robustness:

1. Cancel each job individually, record the exit status individually.
2. Cancel them all in one go, then request the status of running jobs to verify they've in fact been cancelled. 
3. Cancel them all, and then if an issue arises get the the status and do correction.

I'm preferential to 2, but it's not a catch all. Though I suppose this issue is mitigated by the fact that every adapter can handle these issues separately. I just wanted to document the thoughts. Some of these can get you into trouble, especially with LSF which seems to have long delays in cancelling jobs within its internal book keeping. I've had jobs hang around on the order of minutes before finally being marked as cancelled.

jsemler:
@FrankD412 -- Option 2 seems like a reasonable start. I think the scheduler adapters are flexible enough that they can be reworked for more complicated use cases if needed.

FrankD412:
@jsemler -- There's actually a glass jaw there that I've come to find. LSF is kind of slow to cancel its jobs so the status isn't reflected on the order of minutes. Not sure how to get around that because that would mean Maestro couldn't continue its exit without waiting first. I don't know if that's an acceptable solution considering it's handled in the background and ultimately it isn't delaying things because we're simply verifying the jobs stop.

FrankD412:
I was working out the finite state machine for the states and I was thinking of making use of an intermediate state. Right now, adapters execute a cancel, and if the cancel reports a success, then mark the steps as `CANCELLED`. However, if I make used of a `CANCELLING` state, that means I can mark it as in the process of cancelling, then update the state of that step once the scheduler reports that the job has been cancelled or has exited. That feels like the most appropriate way to handle a cancel, that makes it so that the conductor doesn't have to wait, or cancel jobs individually, etc.

FrankD412:
Alright, to move this forward, I'm wrapping some code up. I've decided for now, that I'm going to create the basic objects and then later they can be worked around or tweaked. For now, I'm adding the basics of just storing the return enumeration for cancellation and adding the extra information to the `SubmissionRecord`. That should get @robinson96 the functionality he needed to be able to probe the return code of local execution.

FrankD412:
Just leaving a note, @jwhite242 is going to run a quick test on SLURM.

jwhite242:
Alright, the batch scripts in the steps don't seem to get populated correctly.  The slurm commands look alright, but the two actual srun ... commands are just the place holders '{0}' and '{1}' currently.

FrankD412:
@jwhite242 -- Just to make sure I'm understanding, are those in the `srun` commands or in the shebang shell specification at the top of the file? That's the most recent place I've seen that syntax.

jwhite242:
the srun commands.  the shebang and the #SBATCH...  bits all look fine

jwhite242:
wait, there's an extra '#!' on the shebang that i hadn't noticed before.

FrankD412:
@jwhite242 -- So the odd thing is that this PR doesn't change any of the substitution mechanics for the SLURM adapter. I almost wonder if it's some of the more recent changes in the develop branch. Can you confirm?

FrankD412:
@jwhite242 -- Oh, scratch that. it's totally in develop. Will fix here in a moment.

FrankD412:
@jwhite242 -- The placeholder issue should be fixed now. I'm trying to track down the duplicate `#!` problem now.

jsemler:
@FrankD412 -- I've tested the most recent changes and the sbatch scripts appear to be generating as expected.

PR #178, opened 2019-03-21
Addition of a Dockerfile for tutorials and ease of trying out.

FrankD412:
Added a Dockerfile to allow someone to just checkout the repository and play with Maestro without touching their system.

TODO: This isn't quite done yet because I want to add to the Quick Start guide a section on how to just build and run the container and mess with it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM ubuntu
 LABEL maintainer="Francesco Di Natale dinatale3@llnl.gov"
 
+RUN apt-get update
+RUN apt-get install -y python python-pip git
 ADD . /maestrowf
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu
+LABEL maintainer="Francesco Di Natale dinatale3@llnl.gov"
+
+RUN apt-get update
+RUN apt-get install -y python python-pip git
```
jsemler: Docker recommends defining `apt-get update` and `apt-get install` as part of the same `RUN` statement to avoid caching issues: https://docs.docker.com/develop/develop-images/dockerfile_best-practices

Issue #180, opened 2019-04-16
LocalScriptAdapter shell definition in batch requires shebang and quotes


koning:
The current LocalScriptAdapter takes an argument that is the shell including the shebang, shell="#!/bin/bash".  If a different shell is required, the shebang must be included when adding the the batch section.  In yaml this means that the variable needs to be quoted or it will be interpreted as a comment:
shell: "#!/bin/tcsh" 
not 
shell = #!/bin/tcsh <-- comment, shell will be empty
Removing the #! requirement and adding this when the script is written will eliminate the quoting requirement.

FrankD412:
Fixed by #181

PR #181, opened 2019-04-17
Take out shebang from shell definition and add it when script is written.

koning:
Fixes issue #180


FrankD412:
@koning -- This uncovered an error that has seemingly not been hindering the execution of Maestro, but for some reason the actual invocation of the command line failed. I'm trying to reproduce the issue locally to see what could be going wrong.

maestrowf/interfaces/script/localscriptadapter.py
```diff
@@ -89,7 +89,7 @@ def _write_script(self, ws_path, step):
             restart_path = os.path.join(ws_path, rname)
 
             with open(restart_path, "w") as script:
-                script.write(self._exec)
+                script.write("#!"+self._exec)
```
FrankD412: So, probably a little bit of a knit pick but I suggest the following:
```
script.write("#!{}".format(self._exec))
```

FrankD412: Actually, I take that back. Probably the best way to handle it is:

```
_ = "#!{}\n\n{}\n".format(self._exec, restart)
```

That gets rid of the excess write in the process.

koning: Ok I changed it to this though:

"#!{0}\n\n{1}\n".format(self._exec, cmd)
and
"#!{0}\n\n{1}\n".format(self._exec, restart)

FrankD412: That's fine -- in this case there's no need to be explicit, but it doesn't hurt anything.

Thanks for updating it.

FrankD412:
huh, it looks like Travis also passed this time. Odd that the invocation failed initially.

Issue #197, opened 2019-09-24
Alternate YAMLSpecification load classmethod


ben-bay:
It would be useful if `YAMLSpecification` had a class method that accepted yaml in string form (instead of a path to a yaml file), as follows:

```
@classmethod
def load_spec_from_str(cls, string):
    # loads python objects from the yaml string, returns a YAMLSpecification
```

If that method existed, `load_specification` could just call it:
```
@classmethod
def load_specification(cls, path):
    with open(path, "r") as f:
        return YAMLSpecification.load_spec_from_str(f.read())
```

FrankD412:
So, l was looking at loading from a string as specified above. It turns out that the `yaml` module does not have a `loads` variant of their parsing API (in a fashion akin to `json` which has a `load` for stream objects and `loads` for string objects). Considering most libraries will support stream objects (like file pointers) by default, I almost wonder if providing an API that takes in a stream is more flexible than limiting to a string. With the structure mentioned in your post, I would have to initialize a `StringIO` object from the string. If I can just assume a stream object from the start, it makes the API more flexible because it means I don't need to add APIs for all types of classes. At that point, it becomes the responsibility of the caller to conform streams.

@ben-bay, does that sound reasonable? It would mean that instead of me taking the string, you would just initialize the stream on your end and pass it to Maestro.

ben-bay:
@FrankD412 that sounds reasonable indeed. It hadn't occurred to me to use streams, but this does seem like the correct place to use them.

FrankD412:
@ben-bay -- Merged. Feel free to pull from develop and use the feature.

PR #198, opened 2019-09-25
Extend the Specification interface to break out loading from streams.

FrankD412:
As mentioned in #197, the `load_specification` method handled loading a specification from a file. The method previously would open the filestream, load the contents and parse to a `Specification` object instance. This method has been broken down into two components now: `load_specification` to load from a file and `load specification_from_stream` to load from raw streams.

This serves two purposes: a. it fulfills the use case in #197 such that Maestro can be used now to programmatically generate specifications and get `Specification` objects from them (in the library use case) and b. it breaks down the larger functionality present to smaller functionality (design principles).

@ben-bay -- Feel free to test/offer feedback here.

### DIFF
--- a/maestrowf/datastructures/yamlspecification.py
+++ b/maestrowf/datastructures/yamlspecification.py
@@ -107,8 +107,9 @@ class YAMLSpecification(Specification):
         """
         Load a study specification.
 
-        :param path: Path to a study specification.
-        :returns: A specification object containing the information from path.
+        :param stream: Raw text stream to study YAML specification data.
+        :returns: A specification object containing the information from the
+        passed stream.
         """
 
         try:
--- a/maestrowf/datastructures/yamlspecification.py
+++ b/maestrowf/datastructures/yamlspecification.py
@@ -30,6 +30,7 @@
 """Module containing all things needed for a YAML Study Specification."""
 
 from copy import deepcopy
+from io import StringIO
 import logging
 import yaml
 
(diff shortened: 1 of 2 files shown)


maestrowf/datastructures/yamlspecification.py
```diff
@@ -95,22 +95,33 @@ def load_specification(cls, path):
...
             raise
 
+        return specification
+
+    @classmethod
+    def load_specification_from_stream(cls, stream):
+        """
+        Load a study specification.
+
+        :param path: Path to a study specification.
```
ben-bay: Update this docstring from `path` to `stream`.

ben-bay: Also update the `returns` line in the same way.

FrankD412: Good catch, corrected.

ben-bay:
Everything here looks good to me now.

maestrowf/datastructures/yamlspecification.py
```diff
@@ -95,22 +95,34 @@ def load_specification(cls, path):
...
+        except AttributeError:
+            logger.warning(
+                "*** PyYAML is using an unsafe version with a known "
+                "load vulnerability. Please upgrade your installation "
+                "to a more recent version! ***")
+            spec = yaml.load(stream)
+
         logger.debug("Loaded specification -- \n%s", spec["description"])
         specification = cls()
-        specification.path = path
```
ben-bay: Sorry, one thing I neglected to comment on yesterday-- do we want to move this line up to line 98? So that the `path` attribute still exists? Right now, it looks as if the `path` attribute is gone, which is breaking some of my code.

FrankD412: Good catch -- I thought I had moved it, but it appears I didn't. Keep an eye out for a quick fix soon.

FrankD412: Another thought -- if you're using the stream form of the API, that variable can't be filled by the method. That opens you up to `YAMLSpecification` instances without a populated path member variable.

ben-bay: Right, in that case it would be `None`, or the responsibility of the person that called `load_specification_from_stream` to populate it?

FrankD412: I'll have it populated to `None` so it's explicit that it doesn't have one. That seems to be the best answer here to me.

FrankD412: There's now a quick fix in `develop`.

ben-bay: Thanks! This appears to fix everything.

Issue #201, opened 2019-10-17
sbatch submit issue


stewarthe6:
The arguments to sbatch are in the wrong order.

Should be:
sbatch -D $(workspace) $(path to script)

FrankD412:
@stewarthe6 was trying to load a Conda environment and because of the out of order `sbatch` parameters, Conda was picking them up as inputs and using the `-D <path>` as arguments.

PR #202, opened 2019-10-17
Moved SLURM parameters to be placed in front of the submitted script.

FrankD412:
This corrects the ordering issue discovered in #201.


FrankD412:
@stewarthe6 -- Feel free to give this a test. It was a simple variable swap so it should work now.

stewarthe6:
I gave it a try and it's working now.

PR #209, opened 2019-12-14
Enhances pgen to take in the environment specified in a study.

FrankD412:
Previously, custom parameter generators were unaware of the environment defined in a study specification. This PR tweaks that so that the `StudyEnvironment` data structure is passed to the pgen function so that variables can be consistently defined between the specification and pgen.

This change enables studies which might rely on constructing a list of items in a directory to generate parameters, study specified defaults for parameters, and implementation/method switching (just to name a few).

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -90,7 +90,7 @@ def cancel_study(args):
     return 0
 
 
-def load_parameter_generator(path, kwargs):
+def load_parameter_generator(path, env, kwargs):
     """
     Import and load custom parameter Python files.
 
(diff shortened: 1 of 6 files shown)


FrankD412:
After talking with @jsemler, she brought up the concern about the objects in the `StudyEnvironment` being read-only. After looking through the classes, I confirmed that at least the `GitDependency` is not read-only and presents a concern that users could change values in the environment class by requesting objects and changing them. This possibility is bad as it breaks the specifications ability to be an accurate representation of the static variables a study requires to run, and so the environment classes should be modified to force a read-only circumstance.

PR #67, opened 2018-01-23
Introduces an LSF ScriptAdapter

FrankD412:
This PR adds an LSF adapter so that Maestro can submit via IBM's LSF scheduler.

PR #215, opened 2020-01-13
Test/interfaces/lsf

FrankD412:
This PR adds an LSF adapter so that Maestro can submit via IBM's LSF scheduler.

The previous PR #67 was stale and with the new developments to `develop`, it risked making the branch unstable for those using it. Instead I've moved to a new branch that's been rebased and updated, and can now be tested and modified without potentially breaking the folks currently using the old branch.


FrankD412:
So far as I can tell, monitoring with Maestro using LSF appears to be working. This PR is scoped to fix the monitoring, but does not fix the construction of the `jsrun` command. A user can submit scripts with hardcoded `jsrun` calls using this PR and have them be monitored as expected. Another PR will fix the `jsrun` construction to enable the use of the `$(LAUNCHER)` moniker.

PR #241, opened 2020-04-02
Fix to add PID to local log names.

### LINKED ISSUE #240: The restart step output will overwrite the cmd output
This was found by a merlin user see LLNL/merlin#197

FrankD412:
This PR should fix #240 -- previously the log name was constructed in a fixed format without a unique identifier. This simply adds the PID to the log names.


koning:
@FrankD412 I think the user was looking for a <step>.restart.out,err file. This will still write out a <step>.<pid>.out instead of an out file with restart in the name. This also increases the total number of output files for restarts done in merlin (default retrys is 5). The solution I used just appends to the <step>.out or <step>.restart.out file.

PR #242, opened 2020-04-08
Refactor to move DAG operations to the back end Conductor

FrankD412:
This PR moves DAG expansion to the back end conductor while also introducing a class API for a `Conductor` class. This allows for more flexibility and makes it easier for user to write their own conductors or to have other tools perform conductor style tasking without having to call Maestro's conductor. The new class sets the precedent for the interaction between the Maestro front end and the `Conductor` back end while hiding technical changes from the end user. It also makes the output right at start a lot less verbose.

I need to go through and add docstrings so I've marked it as WIP, but this PR is ready for testing.


maestrowf/abstracts/__init__.py
```diff
@@ -33,23 +33,55 @@
...
 
+LOGGER = logging.getLogger(__name__)
+
+
+class PickleInterface:
+    """A class that implements a general pickle interface using dill."""
+    @classmethod
+    def unpickle(cls, path):
+        """
+        Load an ExecutionGraph instance from a pickle file.
```
jsemler, line 60: `unpickle` seems like it can unpickle any instance as long as the instance matches the passed in `cls`.

maestrowf/abstracts/__init__.py
```diff
@@ -33,23 +33,55 @@
...
 
-__all__ = ("abstractclassmethod", "Dependency", "Graph", "SimObject",
+__all__ = ("abstractclassmethod", "Dependency", "Graph", "PickleInterface",
            "Singleton", "Source", "Specification", "Substitution")
 
+LOGGER = logging.getLogger(__name__)
+
+
+class PickleInterface:
+    """A class that implements a general pickle interface using dill."""
```
jsemler, line 56: This class seems intended to be used as a mixin. If so, consider updating the docstring to clarify.

FrankD412: We talked offline, but I updated the doc.

maestrowf/abstracts/__init__.py
FrankD412: Yeah, this was a direct port to a mixin, updated.

PR #247, opened 2020-04-11
Bugfix for logging that didn't appear in submodules

### LINKED ISSUE #135: Port logger to util library
The logging set up in the maestro.py script is not easily portable, due to its reliance on the globals defined there. It would be beneficial to have access to standardized, maestro-compliant logging that could be imported by other packages, such that processes calling maestro or called by maestro could write compatible logs.

### LINKED ISSUE #246: possible logger bug
When I was working with logger messages in `maestrowf/datastructures/yamlspecification.py`, the logger appeared to be broken. I was only able to view my `logger.warning` with a print.

FrankD412:
Adds a logging utility class and correctly sets the level of the logging facilities so that logging is consistently applied across all submodules. Closes #246 and closes #135 

@ben-bay -- Please try and apply these to your PR locally as a test and see if this fixes your issue in #243.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -29,7 +29,6 @@
 
 """A script for launching a YAML study specification."""
 from argparse import ArgumentParser, ArgumentError, RawTextHelpFormatter
-import inspect
 import logging
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


FrankD412:
It also looks like this PR follows through on the suggestions in #135.

ben-bay:
I tested this and it appears to work, both with `logger.warn` and `logger.warning`. Good job

PR #250, opened 2020-04-18
Addition of an alternative gravity example.

FrankD412:
Addition of an alternative example that illustrates how to set up parameterized studies and associated documentation.


FrankD412:
Closing this PR -- something like this is better suited for the Sheetmusic repository.

PR #196, opened 2019-09-23
Merge current version of flux-dev into develop.

FrankD412:
The `bugfix/flux-dev` has been out in the weeds for a little while. It's been a specific branch for a specific application that contains a lot of ad hoc and as-needed fixes. This PR pulls the current `flux-dev` version of the `FluxScriptAdapter` into `develop` and is intended to bring it up to speed on the current paradigm of using the `Record` classes to pass information. This PR is also intended to clean it up, and now that there's a new version of flux to update it.

PR #251, opened 2020-04-20
Updates to the FluxScriptAdapter to use the latest Flux.

FrankD412:
This PR takes over the old PR #196 that was merging the flux-dev branch into develop.


Jmast:
Issues found with flux:

1. removing procs from study step results in processors being an empty string and causing exception in processors > 0 check in fluxscriptadaptory.py
2. if cores per task is removed from study step, then -c is placed on flux mini run but without any argument
3. if nodes=1, procs=1, cores per task=2 then flux mini run study step works, but if nodes=2, then flux mini fails even though the generated flux script runs fine in flux proxy.
4. there is no stdout/stderr output from the flux mini run process (e.g. like slurm .out and .err output)
5. more error diagnostics needed for flux failures in general to help diagnose issues

Jmast:
nodes: 4
        procs: 8
        cores per task: 18

I would expect this to run 2 tasks per node on the 4 nodes with each task using 18 cores
This seems to work with flux:
    flux mini run -N 4 -n 8 -c 18 hostname
but in context of maestro, it seems to be an unallocatable request.

FrankD412:
> ```
>     nodes: 4
>     procs: 8
>     cores per task: 18
> ```
> 
> I would expect this to run 2 tasks per node on the 4 nodes with each task using 18 cores
> This seems to work with flux:
> flux mini run -N 4 -n 8 -c 18 hostname
> but in context of maestro, it seems to be an unallocatable request.

@Jmast -- has the above been resolved? I recently made some changes to the adapter to use Flux's built-in nesting. Hoping I fixed this issue.

PR #259, opened 2020-05-13
Addition of a simple dry-run capability.

### LINKED ISSUE #228: Enable a dry run feature that will construct the study workspace, but not execute a study.
Users have expressed that a nice to have feature would be a dry run mode where the workspace is constructed but the study is not executed. This feature would allow users to see sample scripts of what would be generated and debug issues independently of having to run a full study.

FrankD412:
This PR adds a simple dry run that forces Maestro to skip scheduler status checking, execution, and generates all workspaces and scripts so that users can see what's generated before kicking off studies. Closes #228.

### DIFF
--- a/maestrowf/abstracts/enums/__init__.py
+++ b/maestrowf/abstracts/enums/__init__.py
@@ -65,6 +65,7 @@ class State(Enum):
     TIMEDOUT = 10
     UNKNOWN = 11
     CANCELLED = 12
+    DRYRUN = 13
 
 
 class StudyStatus(Enum):
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -302,7 +302,7 @@ class ExecutionGraph(DAG, PickleInterface):
     """
 
     def __init__(self, submission_attempts=1, submission_throttle=0,
-                 use_tmp=False):
+                 use_tmp=False, dry_run=False):
         """
         Initialize a new instance of an ExecutionGraph.
 
@@ -336,6 +336,7 @@ class ExecutionGraph(DAG, PickleInterface):
         # throttling, etc. should be listed here.
         self._submission_attempts = submission_attempts
         self._submission_throttle = submission_throttle
+        self.dry_run = dry_run
 
         # A map that tracks the dependencies of a step.
         # NOTE: I don't know how performant the Python dict structure is, but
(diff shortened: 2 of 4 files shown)


ben-bay:
Does this look like `maestro dryrun hello.yaml` at the command line?

FrankD412:
Oh, silly me -- I should have posted how to actually test this @ben-bay. Good call. See below. I settled on the `--dryrun` flag since that has been in Maestro for a while, but it simply raised a `NotImplemented` exception.

`maestro run --dryrun <path to spec>` -- There's a SLURM version of the LULESH spec in the `samples` directory which I tested with. Feel free to throw your own specs at it. It should still acquire the dependencies as normal so those should error out as expected.

The output should look as follows:
```
Step Name                     Workspace          State    Run Time    Elapsed Time    Start Time    Submit Time    End Time                      Number Restarts
----------------------------  -----------------  -------  ----------  --------------  ------------  -------------  --------------------------  -----------------
run-lulesh_ITER.200.SIZE.200  ITER.200.SIZE.200  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.529221                  0
run-lulesh_ITER.300.SIZE.200  ITER.300.SIZE.200  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.538404                  0
run-lulesh_ITER.300.SIZE.300  ITER.300.SIZE.300  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.560271                  0
run-lulesh_ITER.100.SIZE.300  ITER.100.SIZE.300  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.545585                  0
run-lulesh_ITER.200.SIZE.100  ITER.200.SIZE.100  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.499471                  0
make-lulesh                   make-lulesh        DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:30.440264                  0
run-lulesh_ITER.100.SIZE.200  ITER.100.SIZE.200  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.516734                  0
run-lulesh_ITER.200.SIZE.300  ITER.200.SIZE.300  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.553485                  0
run-lulesh_ITER.300.SIZE.100  ITER.300.SIZE.100  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.508730                  0
run-lulesh_ITER.100.SIZE.100  ITER.100.SIZE.100  DRYRUN   --:--:--    --:--:--        --            --             2020-05-13 08:49:31.491691                  0
```

ben-bay:
The nomenclature in merlin for this is `merlin run --dry`, I chose that because it felt redundant to say "run" twice.

FrankD412:
It's funny that you say that, I was typing up my comment and I was thinking the exact same thing.

FrankD412:
Alright, update -- `maestro run --dry <spec>` is the new command.

ben-bay:
This works for me. Maybe instead of asking:
```
Would you like to launch the study? [yn]
```
You could re-word it so it's clear that the actual study isn't being launched. Even though above there is already the dry run bool, this might be helpful

jwhite242:
> This works for me. Maybe instead of asking:
> 
> ```
> Would you like to launch the study? [yn]
> ```
> 
> You could re-word it so it's clear that the actual study isn't being launched. Even though above there is already the dry run bool, this might be helpful

for the dry run, maybe just use the dry run flag to skip the question entirely?  Not sure a message is really needed there.

FrankD412:
@ben-bay @jsemler -- Does it make sense here to just skip the confirmation? I think it would be alright, since the dry run should be light weight.

ben-bay:
> @ben-bay @jsemler -- Does it make sense here to just skip the confirmation? I think it would be alright, since the dry run should be light weight.

Yup

FrankD412:
Cool -- auto yes is enabled by default for dry runs.

PR #262, opened 2020-05-13
enabled raw sbatch errors to be logged

crkrenn:
Minor change that will enable easier debugging of slurm batch errors.

### DIFF
--- a/maestrowf/interfaces/script/slurmscriptadapter.py
+++ b/maestrowf/interfaces/script/slurmscriptadapter.py
@@ -200,7 +200,8 @@ class SlurmScriptAdapter(SchedulerScriptAdapter):
             jid = re.search('[0-9]+', output).group(0)
             return SubmissionRecord(SubmissionCode.OK, retcode, jid)
         else:
-            LOGGER.warning("Submission returned an error.")
+            LOGGER.warning("Submission returned an error (see next line).")
+            LOGGER.warning(output)
             return SubmissionRecord(SubmissionCode.ERROR, retcode)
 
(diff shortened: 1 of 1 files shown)


maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -200,7 +200,8 @@ def submit(self, step, path, cwd, job_map=None, env=None):
             jid = re.search('[0-9]+', output).group(0)
             return SubmissionRecord(SubmissionCode.OK, retcode, jid)
         else:
-            LOGGER.warning("Submission returned an error.")
+            LOGGER.warning("Submission returned an error (see next line).")
+            LOGGER.warning(output)
```
FrankD412, line 204: @crkrenn -- Two things:
1. Does `sbatch` really return error messages on `stdout`? It'd be surprised if it ends up on `stdout` normally since `stderr` is returned there by default. Would be worth testing to make sure that `output` here does contain the error if you haven't already. If you have, then just disregard this item. :)
2. It's a knitpick, but it's cleaner to do single log calls:
    ```
    LOGGER.warning("Submission returned an error (see next line).\n%s", output) # or err depending
    ```

Lastly, thank you for contributing. The help is appreciated. 👍

crkrenn: I made the changes Frank suggested.

FrankD412: Appreciated. The style check is still failing due to some whitespace. Looks like it's on L204. I'll merge it in after that passes.

Issue #234, opened 2020-03-12
#SBATCH -n header to specify number of processors per node


srcopela:
I'm encountering a situation where one of the HPC systems I run on is configured such that only one processor is allocated per node unless explicitly otherwise specified in the slurm script.  Concretely, the auto-generated *.sh file from maestro looks something like:
 
#!/bin/bash
#SBATCH --comment "Run the simulation."
#SBATCH -J run-name
#SBATCH -p pbatch
#SBATCH -N 1
#SBATCH -A science
#SBATCH -t 00:05:00
 
srun –n 2 –N 1 /usr/gapps/code infile
 
Because  the #SBATCH block only specifies  #SBATCH -N 1 and doesn’t specify  the number of processors with #SBATCH -n 2, the machine defaults to allocating the single node and 1 processor. So when the srun runs it tries to run on the single node with 2 processors and fails because SLURM only allocated a single processor.
 
To fix this we’d need to be able to specify the #SBATCH -n 2 in the sbatch script to allocate more than one processor per node.
 
#!/bin/bash
#SBATCH --comment "Run the simulation."
#SBATCH -J run-name
#SBATCH -p pbatch
#SBATCH -N 1
#SBATCH -n 2
#SBATCH -A science
#SBATCH -t 00:05:00
 
srun –n 2 –N 1 /usr/gapps/code infile
 

The current workaround for me is to include the #SBATCH -n  <nprocs> line as the first entry of the cmd field for the step, but it would be nice if maestro handled that behavior in a sane way.

Thanks!

FrankD412:
So where this gets tricky is that in my experience most clusters don't expressly require specifying the number of processors in the header. I've been thinking about this over the last day or two and I think that the right solution might be a setting in the `batch` section of the spec that toggle enforcing the processors in the header. Something like `header_procs` or `enforce_procs`? That would be set to `true` or `false`.

jwhite242:
What about just using presence of the 'procs' key and putting it in the header if found; even with the boolean flag you proposed you'd still need to tell it how many procs there are.

PR #263, opened 2020-05-16
Tweaks and fixes to the SLURM header.

### LINKED ISSUE #239: Slurm reservation and bank interactions
Presently, Maestro's SLURM adapter allows the use of a reservation (good) and a specified bank (good), but does not allow these two settings to be used together; i.e., I specify reservation myawesomereservation and bank thebankiwanttocharge, the SBATCH scripts are constructed such that get the reservation myawesomereservation, but the bank is not filled in and I just get my default bank.

Can this be fixed?

The specific snippet of interest is in get_header() in interfaces/script/slurmscriptadapter.py:

`        modified_header = [self._exec]
        for key, value in self._header.items():
            # If we're looking at the bank and the reservation header exists,
            # skip the bank to prefer the reservation.
            if key == "bank" and "reservation" in self._batch:
                if self._batch["reservation"]:
                    continue
            modified_header.append(value.format(**batch_header))

        return "\n".join(modified_header)` 

There might need to be more sophisticated handling here.

FrankD412:
Adds the ability to specify GPUs, fixes reservation pairing with accounts, and now uses a ChainMap to handle the internal key conflicts between the batch and step settings. Also introduces the exclusive key. Changed to full parameter names for clarity.

Closes #239. Might be useful for #234.

### DIFF
--- a/maestrowf/datastructures/schemas.json
+++ b/maestrowf/datastructures/schemas.json
@@ -42,7 +42,8 @@
           "gpus": {"type": "integer"},
           "cores per task": {"type": "integer"},
           "walltime": {"type": "string", "minLength": 1},
-          "reservation": {"type": "string", "minLength": 1}
+          "reservation": {"type": "string", "minLength": 1},
+          "exclusive": {"type": "boolean"}
         },
         "additionalProperties": false,
         "required": [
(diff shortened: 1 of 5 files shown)


FrankD412:
@tadesautels -- This change appears to work in the normal case, but I haven't been able to test a reservation. That use case isn't exercised too often so far as I know, so I'll merge this. If you test this and find it to not work, we can revisit.

@srcopela -- This should fix your inability to add `ntasks` to the header. If it's present in the `batch` section, Maestro will add it to the header. Otherwise, if not, it will only place it into the `srun` command.

PR #265, opened 2020-05-17
modified executiongraph to round datetimes to nearest second

crkrenn:
The main reason to do this was to improve the formatting of `maestro status`. 

An alternate solution would be to modify the getters, or status read and write, but this solution is a little cleaner and faster, and it has the same result. 

If there is a use case that needs more precise datetime stamps, other solutions could be considered. 

And, to give credit where credit is due: 

https://stackoverflow.com/questions/47792242/rounding-time-off-to-the-nearest-second-python.

### DIFF
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -14,7 +14,7 @@ from maestrowf.abstracts.enums import JobStatusCode, State, SubmissionCode, \
 from maestrowf.datastructures.dag import DAG
 from maestrowf.datastructures.environment import Variable
 from maestrowf.interfaces import ScriptAdapterFactory
-from maestrowf.utils import create_parentdir, get_duration
+from maestrowf.utils import create_parentdir, get_duration, round_datetime_seconds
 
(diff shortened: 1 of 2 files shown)


maestrowf/datastructures/core/executiongraph.py
```diff
@@ -14,7 +14,7 @@
 from maestrowf.datastructures.dag import DAG
 from maestrowf.datastructures.environment import Variable
 from maestrowf.interfaces import ScriptAdapterFactory
-from maestrowf.utils import create_parentdir, get_duration
+from maestrowf.utils import create_parentdir, get_duration, round_datetime_seconds
```
FrankD412, line 17: It looks to me this is the only style error. Otherwise, could you put the stackoverflow link in the comments? You can split the link between two lines to fit in the character limit.

FrankD412:
@crkrenn -- In the interest of getting this into the release, I'll be merging it in without the style fixes. It's technically bad practice, but I'd like to release 1.1.7 on Monday.

crkrenn:
Thanks for the help!

PR #266, opened 2020-05-17
Adds a check for UNKNOWN state.

### LINKED ISSUE #264: Tasks never complete due to SLURM cache clearing too quickly.
In some instances when the SLURM job cache is behaving erratically or removes completed jobs from the `squeue` cache too quickly, it leaves finishing jobs in an UNKNOWN state, resulting in the state being left in RUNNING. The appropriate guess here for the FINISHING state would be to assume the job finished and set it accordingly.

Found when testing #263

FrankD412:
Fixes #264 -- After some testing it appears SLURM loses track of the jobs when in completing state because the system runs out of memory and aborts them. I think it might be reasonable to assume that if you can't find the status, then it likely was killed for some abnormal reason as I do have jobs that seem to successfully finished.

Previously: If SLURM is not caching state properly, but steps
are found to be in FINISHING previous state when seen as UNKNOWN,
mark them complete. Otherwise, mark it unknown, and let the next
step run and fail if the UNKNOWN step failed.

### DIFF
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -776,18 +776,19 @@ class ExecutionGraph(DAG, PickleInterface):
                             "Step '%s' marked as finished. Found to be in "
                             "UNKNOWN state when previously in FINISHING.",
                             " Adding to completed set.", name)
+                        self.completed_steps.add(name)
                     else:
-                        # We don't know the current state, and previously it
-                        # might have finished successfully. We'll mark it
-                        # unknown, and let the next step fail if this errored.
-                        record.mark_end(State.UNKNOWN)
+                        # We don't know the current state, so we mark it as
+                        # failed. It appears that if something happens in the
+                        # completion hooks, a scheduler can abort.
+                        record.mark_end(State.FAILED)
                         LOGGER.info(
                             "Step '%s' found in UNKNOWN state. Step was found "
                             "in '%s' state previously, making as UNKNOWN. "
                             "Adding to completed step in the event that step "
-                            "succeeded. Adding to completed set.",
+                            "succeeded. Adding to failed set.",
                             name, record.status)
-                    self.completed_steps.add(name)
+                        cleanup_steps.update(self.bfs_subtree(name)[0])
                     self.in_progress.remove(name)
 
                 elif status == State.CANCELLED:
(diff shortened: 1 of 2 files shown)


FrankD412:
After some digging around, in SLURM's case, when a job goes off the rails while it's completing or in some intermediary state it's not reported as failed. I think SLURM wipes it from the job list entirely on exit when it gets aborted. I've made it to so that UNKNOWN is treated like failed, except it'll make the step as unknown (and will still mark the dependent steps as failed due to UNKNOWN).

jwhite242:
> After some digging around, in SLURM's case, when a job goes off the rails while it's completing or in some intermediary state it's not reported as failed. I think SLURM wipes it from the job list entirely on exit when it gets aborted. I've made it to so that UNKNOWN is treated like failed, except it'll make the step as unknown (and will still mark the dependent steps as failed due to UNKNOWN).

Anything interesting show up in the .out/.err that the subprocess captures, or better yet from the slurm .out file?

FrankD412:
> > After some digging around, in SLURM's case, when a job goes off the rails while it's completing or in some intermediary state it's not reported as failed. I think SLURM wipes it from the job list entirely on exit when it gets aborted. I've made it to so that UNKNOWN is treated like failed, except it'll make the step as unknown (and will still mark the dependent steps as failed due to UNKNOWN).
> 
> Anything interesting show up in the .out/.err that the subprocess captures, or better yet from the slurm .out file?

I've closed my laptop, but at I believe that the error was caught in the .err file that was generated by SLURM as specified in the header. That how I found out it went off the rails.

jwhite242:
Would it make sense to log the slurm .err that the subprocess captures along with this info message, assuming it was populated?

Also, regarding the PR, this looks like it just marks any unknown state job as failed instead of checking the previous state and marking them completed if appropriate?

FrankD412:
> Would it make sense to log the slurm .err that the subprocess captures along with this info message, assuming it was populated?

I don't know how long this error output can be -- the info is certainly there using the wrappers on LC, but `squeue` is unaware of the info (at least I haven't found a way to get it out of that command yet). It is automatically placed into the .err file as specified by the SLURM adapter.
 
> Also, regarding the PR, this looks like it just marks any unknown state job as failed instead of checking the previous state and marking them completed if appropriate?

Let me edit that, I didn't go back and change it. I had originally coded it that way because I thought SLURM was losing track of it. But as it turns out, it ran out of memory in the completion stage and erred out. I had intended on setting the state to UNKNOWN, but it looks like clean up forces it to FAILED.

PR #271, opened 2020-05-19
Feature/crkrenn/multidimensional sampling

crkrenn:
I've implemented three "NTNF" style sampling interfaces. 

Because this is of general interest and can be implemented in maestro alone, I am submitting it here. 

Best_candidate sampling is next.

PR #272, opened 2020-05-19
Feature/crkrenn/best candidate sampling

crkrenn:
@jsemler @FrankD412, 

I've implemented a "best candidate" sampling method based on coding by Cody Raskin (@antimatterhorn). See https://bl.ocks.org/mbostock/6224396 for a related visualization.

I've put this in a separate pull request because it requires pandas, numpy, & scipy packages to run. 

What is your advice regarding the extra packages? I've installed them in my own venv, and the best candidate sampling code works. But unless I modify requirements.txt, the sampling code won't work out of the box for other users.  

-Chris

### DIFF
--- a/samples/multidimensional_sampling/pgen_sample.py
+++ b/samples/multidimensional_sampling/pgen_sample.py
@@ -1,5 +1,7 @@
 """This file implements several sampling methods"""
 
+# original best candidate coding done by Cody Raskin in 2019.
+
 import random
 import logging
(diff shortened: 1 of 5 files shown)


FrankD412:
@crkrenn -- @jwhite242 and I were talking and we think that these (and #271) would be useful for a sampling section right after the parameter documentation Jeremy just got done writing. What do you think?

PR #275, opened 2020-05-21
Pgen docs

jwhite242:
Adds documentation of the parameter generators


docs/source/parameters.rst
```diff
@@ -0,0 +1,228 @@
...
+
+Pgen Arguments
+**************
+
+There is an additional pgen feature that can be used to make them more dynamic.  The above example generates a fixed set of parameters, requiring editing the itertools_pgen.py file to change that.  Maestro supports passing arguments to these generator functions on the command line:
+
+
+.. code-block:: bash
+
+   $ maestro run study.yaml --pgen itertools_pgen_pargs.py --parg "SIZE_MIN:10" --parg "SIZE_STEP: 10" --parg "NUM_SIZES:4"
```
FrankD412, line 112: There's a space in one of the args, I think the space won't be stripped out.

jwhite242: whoops, good catch

FrankD412:
@jwhite242 -- Are you still working on other examples?

jwhite242:
> @jwhite242 -- Are you still working on other examples? What I saw looked pretty good.

So, was thinking no, but just realize I've not put in any discussion of the env block being accessible in the pgen, so maybe one more for that?

FrankD412:
> > @jwhite242 -- Are you still working on other examples? What I saw looked pretty good.
> 
> So, was thinking no, but just realize I've not put in any discussion of the env block being accessible in the pgen, so maybe one more for that?

Yeah, did you want to carry it forward with a sampling that has a picture?

FrankD412:
@jwhite242 -- this is the current example for a [random sample LULESH spec](https://github.com/LLNL/maestrowf/blob/develop/samples/parameterization/lulesh_montecarlo_args.py).

jwhite242:
> @jwhite242 -- this is the current example for a [random sample LULESH spec](https://github.com/LLNL/maestrowf/blob/develop/samples/parameterization/lulesh_montecarlo_args.py).

That's perfect demo of it.  Added that into that section.  Did also add an extra parameter, which may be of interest: seed.  I can add that to the sample generator python file as well if you want.  Maybe want some tweaks there, but if the seed isn't in the yaml file, it'll behave as it did before, with None passed into the seed function being the same as the default behavior of using the system time.

jwhite242:
Also, since you highlighted one of the built in samples, what about adding a samples sub section to so some automated listing of those in the docs to make them more accessible without diving into the code?

FrankD412:
> > @jwhite242 -- this is the current example for a [random sample LULESH spec](https://github.com/LLNL/maestrowf/blob/develop/samples/parameterization/lulesh_montecarlo_args.py).
> 
> That's perfect demo of it. Added that into that section. Did also add an extra parameter, which may be of interest: seed. I can add that to the sample generator python file as well if you want. Maybe want some tweaks there, but if the seed isn't in the yaml file, it'll behave as it did before, with None passed into the seed function being the same as the default behavior of using the system time.

If you don't mind adding it to the spec that'd be awesome. I'd like to keep it consistent -- even if it is commented out, I think it's beneficial to have that visual connection so someone isn't confused.


> Also, since you highlighted one of the built in samples, what about adding a samples sub section to so some automated listing of those in the docs to make them more accessible without diving into the code?

Sure, I'll have to look and see if there's a way to scan that directory in Sphinx. Do you know of a way already? Otherwise we could save that for another PR. Also see #271 and #272 -- I recommended that maybe those could be put into docs (I think I commented on #271). Thoughts?

jwhite242:
> > > @jwhite242 -- this is the current example for a [random sample LULESH spec](https://github.com/LLNL/maestrowf/blob/develop/samples/parameterization/lulesh_montecarlo_args.py).
> > 
> > 
> > That's perfect demo of it. Added that into that section. Did also add an extra parameter, which may be of interest: seed. I can add that to the sample generator python file as well if you want. Maybe want some tweaks there, but if the seed isn't in the yaml file, it'll behave as it did before, with None passed into the seed function being the same as the default behavior of using the system time.
> 
> If you don't mind adding it to the spec that'd be awesome. I'd like to keep it consistent -- even if it is commented out, I think it's beneficial to have that visual connection so someone isn't confused.
> 
> > Also, since you highlighted one of the built in samples, what about adding a samples sub section to so some automated listing of those in the docs to make them more accessible without diving into the code?
> 
> Sure, I'll have to look and see if there's a way to scan that directory in Sphinx. Do you know of a way already? Otherwise we could save that for another PR. Also see #271 and #272 -- I recommended that maybe those could be put into docs (I think I commented on #271). Thoughts?

Was thinking just using a bunch of literalincludes and starting the convention that new samples get put into the docs.  Not sure if there is an existing way to automate it out side of writing a custom directive as my quick google searching didn't turn up much.

Looking at those other two PRs, those are certainly some interesting case studies on using pgen to build up complicated sampling tools.  Would be interesting to build up some in depth examples in the docs with things like this (having a spec with some example code that uses it for real work would be really cool with those sampling tools).  But, I'm guessing that the intent of the PR's isn't just to show them off in the docs, rather making them usable by default?   The extra dependencies in there do make for a potential headache, plus the maintenance burden of the sampling algorithms and the precedent likely leading to many more sampling PRs that push this tool into the sampling space too.  Seems like a separate tool that would benefit from it's own update cycle rather than tying it to maestro's code versions

FrankD412:
> Looking at those other two PRs, those are certainly some interesting case studies on using pgen to build up complicated sampling tools. Would be interesting to build up some in depth examples in the docs with things like this (having a spec with some example code that uses it for real work would be really cool with those sampling tools). But, I'm guessing that the intent of the PR's isn't just to show them off in the docs, rather making them usable by default? The extra dependencies in there do make for a potential headache, plus the maintenance burden of the sampling algorithms and the precedent likely leading to many more sampling PRs that push this tool into the sampling space too. Seems like a separate tool that would benefit from it's own update cycle rather than tying it to maestro's code versions

That would probably be the better place for them. I'm not sure the appropriate place for those right now -- we could work on starting a new package for them.

> Was thinking just using a bunch of literalincludes and starting the convention that new samples get put into the docs. Not sure if there is an existing way to automate it out side of writing a custom directive as my quick google searching didn't turn up much.

Works for me!

docs/source/parameters.rst
```diff
@@ -0,0 +1,477 @@
...
+    
+       return p_gen
+
+
+The object simply builds the same nested key:value pairs seen in the global.parameters block available in the yaml specification.
+
+For this simple example above, this may not offer compelling advantages over writing out the flattened list in the yaml specification directly.  This programmatic approach becomes preferable when expanding studies to use hundreds of parameters and parameter values or requiring non-trivial parameter value distributions.  The following examples will demonstrate these scenarios using both standard python library tools and additional 3rd party packages from the larger python ecosystem.
+
+EXAMPLES:
+  find another sampling algorithm: latin hypercube, or something else from scikit-learn? what about stats models?
```
FrankD412, line 230: Is this still and outstanding todo item?

FrankD412:
@jwhite242 -- I threw up a [PR](https://github.com/jwhite242/maestrowf/pull/1) on your fork.

docs/source/parameters.rst
```diff
@@ -0,0 +1,373 @@
...
+
+   global.parameters:
+       INPUT:
+           values  : [input1.in, input2.in, input3.in, input1.in, input2.in, input3.in]
+           label   : INPUT.%%
+       VERSION:
+           values  : [4.0.0, 4.0.0, 4.0.0, 5.0.0, 5.0.0, 5.0.0]
+           label   : VERSION.%%
+
+This example parameterizes the inputs and the version of the code being run.  Maestro will run each input with the different code version.  The above example assumes that all the code versions share a base path, **$(CODE_PATH)** which is inserted via the token replacment mechanism to yeild the full paths (e.g. /usr/gapps/code/4.0.0/code.exe).
```
FrankD412, line 135: I feel like paths and references to YAML specific components: `$(CODE_PATH)` is more easily visible as a YAML component.

docs/source/parameters.rst
```diff
@@ -0,0 +1,373 @@
...
+
+.. image:: pgen_images/cheb_map.png
+
+
+Referencing Values from a Specification's Env Block
+===================================================
+
+In addition to command line arguments via :ref:`pargs_section`, the variables defined in the **env** block in the workflow specification file can be accessed inside the :py:class:`~maestrowf.datastructures.core.ParameterGenerator` objects, which is passed in to **get_custom_generator** as the first argument.  The lulesh sample specification can be extended to store the default values for the **pgen**, enhancing the reproducability of the generator.  The following example makes use of an optional **seed** parameter which can be added to the **env** block or set via **pargs** to make a repeatable study specification, while omitting it can enable fully randomized workflows upon every instantiation.  The variables are accessed via the :py:class:`~maestrowf.datastructures.core.StudyEnvironment`'s :py:func:`~maestrowf.datastructures.core.StudyEnvironment.find()` function, which will return ``None`` if the variable is not defined in the study specification.
```
FrankD412, line 367: I think when referring to sections the ** bolding works just fine. @jwhite242 -- what do you think?

docs/source/parameters.rst
```diff
@@ -0,0 +1,477 @@
...
+       depends: [make-lulesh]
+
+Even though this is defined in Maestro as a single step, Maestro will automatically run this step with each parameter combinations. This makes it very easy to setup studies and apply parameters to be run.
+
+.. note::
+
+   Maestro will only use parameters if they've been defined in at least one step
+
+.. note
+   add example with using the label in steps too
```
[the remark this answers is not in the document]
jwhite242: oh, hah, yeah.. reading that prompted the extra example, and then was apparently wiped from memory..  will remove

docs/source/parameters.rst
```diff
@@ -0,0 +1,373 @@
...
+           $(LULESH)/lulesh2.0 -s $(SIZE) -i $(ITERATIONS) -p > $(outfile)
+       depends: [make-lulesh]
+
+Even though this is defined in Maestro as a single step, Maestro will automatically run this step with each parameter combinations. This makes it very easy to setup studies and apply parameters to be run.
+
+.. note::
+
+   Maestro will only use parameters if they've been defined in at least one step
+
+In addition to direct access to parameter values, parameter labels can also be used in steps:
```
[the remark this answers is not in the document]
jwhite242: Definitely like that better

docs/source/parameters.rst
jwhite242: meaning revert the in-text mentions of monikers to be ``$(CODE_PATH)`` instead of the bolded format here?

docs/source/parameters.rst
jwhite242: If you mean the :ref:`pargs_section`, that's a link, not just formatting.  And following from the previous comment, is the suggestion here that any key/block name that can be in a YAML specification should be formatted as ` ``item`` ` instead of ``**item**``?, so seed and env would get the former, and pgen, pargs, get_custom_generator would be the latter? (mixed feelings on the last one not being formatted as code, but could go either way)

docs/source/parameters.rst
FrankD412: I think that would look better. What do you think?

docs/source/parameters.rst
FrankD412: I think that `get_custom_generator` and `seed` look better. **pargs** is a command line option, so I think that could go either way. Out of curiosity, is **pgen** bolded to match **pargs**? -- I think those could be put into tick marks (I think that would match the style elsewhere).

PR #277, opened 2020-05-28
Replace unicode quote with ascii quote

jwhite242:
Fix stray unicode quote causing import issues in python versions < 3.7


jwhite242:
so only concern i have with this PR is why didn't any of the tests choke on this unicode character.  travis and/or that xenial ubuntu image setting a utf8 friendly locale or environment variable?

FrankD412:
@jwhite242 -- Yeah, that's what I was thinking about. All the images have to `pip install` Maestro before running LULESH. I would imagine that setuptools would have erred out, so not sure.

PR #298, opened 2020-07-14
Update parameters.rst

tanimislam:
Make as explicit as possible that if you run with PGEN, then just comment out the global.parameters block in your YAML spec file.

### DIFF
--- a/docs/source/parameters.rst
+++ b/docs/source/parameters.rst
@@ -55,7 +55,6 @@ Defining multiple parameters in the parameter block will share a 1:1 mapping. Ma
 Maestro does not do any additional operations on parameters such as cross products. If more complex methodologies are required to define parameters then the use of Maestro's :ref:`pgen_section` is recommended.
 
 .. note
-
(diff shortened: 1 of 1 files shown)


docs/source/parameters.rst
```diff
@@ -54,6 +54,9 @@ Defining multiple parameters in the parameter block will share a 1:1 mapping. Ma
 
 Maestro does not do any additional operations on parameters such as cross products. If more complex methodologies are required to define parameters then the use of Maestro's :ref:`pgen_section` is recommended.
 
+.. note
```
FrankD412, line 57: A couple of things:
- The array length error isn't the only error that can be thrown. Generally, Maestro will still validate a specification so other validation errors might appear.
- Do you mind rephrasing some of this to reflect the above bullet? Something like: 
```
Even when using the pgen functionality from the command line, Maestro will still verify the provided specification is valid as if it planned to use it entirely. If you are using the `global.parameters` block solely as documentation, we recommend that you comment out the block allowing the validator to ignore it. 
```

Does that sound reasonable to you? Also, I'm not 100% sure that's the best place to put the note when looking through the documentation. I think it might be better in the actual pgen section as a caveat -- since then they'll know what pgen is.

tanimislam: Hello, I made the changes you suggested.

For point 2, this is an obvious trip-up. At least here (in the PGEN section) possible subtlety with Cartesian product is identified, and it is obvious what to look for.

PR #303, opened 2020-08-17
Dagvis

jwhite242:
Add visualization option for the workflow's dag, including multiple output formats (matplotlib, dot file, graphml) and two layout options (spring layout and dot hierarchical).

### DIFF
--- a/maestrowf/datastructures/core/study.py
+++ b/maestrowf/datastructures/core/study.py
@@ -68,6 +68,7 @@ class StudyStep:
     def __init__(self):
         """Object that represents a single workflow step."""
         self.name = ""
+        self.base_name = ""     # Stores unparameterized name
         self.description = ""
         self.run = {
                         "cmd":              "",
@@ -83,6 +84,9 @@ class StudyStep:
                         "reservation":      ""
                     }
 
+        self._param_vals = {}   # Better to be None?
+        self._param_labels = {}
+
     def apply_parameters(self, combo):
         """
         Apply a parameter combination to the StudyStep.
@@ -92,8 +96,13 @@ class StudyStep:
         """
         # Create a new StudyStep and populate it with substituted values.
         tmp = StudyStep()
+
+        base_name = tmp.name
         tmp.__dict__ = apply_function(self.__dict__, combo.apply)
         # Return if the new step is modified and the step itself.
+        tmp.base_name = base_name  # Why doesn't this work here?
+        tmp._param_vals = combo.param_vals
+        tmp._param_labels = combo.param_labels
 
         return self.__ne__(tmp), tmp
 
@@ -121,6 +130,20 @@ class StudyStep:
         : returns: True if other is not equal to self, False otherwise.
         """
         return not self.__eq__(other)
+
+    @property
+    def param_vals(self):
+        """
+        Return dict of parameter values for this step
+        """
+        return self._param_vals
+
+    @property
+    def param_labels(self):
+        """
+        Return dict of parameter labels for this step
+        """
+        return self._param_labels
 
 
 class Study(DAG, PickleInterface):
@@ -164,7 +187,8 @@ class Study(DAG, PickleInterface):
     """
 
     def __init__(self, name, description,
-                 studyenv=None, parameters=None, steps=None, out_path="./"):
+                 studyenv=None, parameters=None, steps=None, out_path="./",
+                 draw=False):
         """
         Study object used to represent the full workflow of a study.
 
@@ -179,6 +203,7 @@ class Study(DAG, PickleInterface):
         :param studyenv: A populated StudyEnvironment instance.
         :param parameters: A populated Parameters instance.
         :param outpath: The path where the output of the study is written.
+        :param draw: Flag controlling generation of dot style visualization
         """
         # The basic study information
         self.name = name
@@ -224,6 +249,9 @@ class Study(DAG, PickleInterface):
             for step in steps:
                 # Deep copy because it prevents modifications after the fact.
                 self.add_step(step)
+
+        # Set dag vis flag
+        self.draw = draw
 
     @property
     def output_path(self):
@@ -446,7 +474,7 @@ class Study(DAG, PickleInterface):
         # Items to store that should be reset.
         LOGGER.info(
             "\n==================================================\n"
-            "Constructing parameter study '%s'\n"
+            "Constructing study '%s'\n"
             "==================================================\n",
             self.name
         )
@@ -644,6 +672,7 @@ class Study(DAG, PickleInterface):
                     self.step_combos[step].add(combo_str)
 
                     modified, step_exp = node.apply_parameters(combo)
+                    step_exp.base_name = step_exp.name
                     step_exp.name = combo_str
 
                     # Substitute workspaces into the combination.
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -207,8 +207,10 @@ def run_study(args):
         parameters = spec.get_parameters()
 
     # Setup the study.
+    LOGGER.debug("Draw flag: {}".format(args.draw))
     study = Study(spec.name, spec.description, studyenv=environment,
-                  parameters=parameters, steps=steps, out_path=output_path)
+                  parameters=parameters, steps=steps, out_path=output_path,
+                  draw=args.draw)
 
     # Check if the submission attempts is greater than 0:
     if args.attempts < 1:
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -44,7 +44,9 @@ def load_readme():
         "coloredlogs",
         "chainmap ; python_version<'3'",
     ],
-    extras_require={},
+    extras_require={"networkx": ["networkx"],
```
FrankD412, line 47: If I'm understanding, this is `extra_key` that maps to a list of `[extra_requirement1, ...]`. Would it make sense to make this something like:

```
extras_require={"vis": ["networkx", "matplotlib", "pygrapyviz"],
}
```

jwhite242: That's a good idea.  Also be good if i fix it so pygraphviz is spelled correctly.

FrankD412: It might be good to document in the installation information how to install the `vis` aspects too.

maestrowf/maestro.py
```diff
@@ -334,6 +336,11 @@ def setup_argparser():
     run.add_argument("--dry", action="store_true", default=False,
                      help="Generate the directory structure and scripts for a "
                      "study but do not launch it. [Default: %(default)s]")
+    run.add_argument("--draw", type=str, action="append", default=[],
```
jsemler, line 339: Are there any thoughts on enabling this feature to be used without having to submit a Maestro run?

jwhite242: that already works

jwhite242: ok, i need to clarify: it works with dry run, so it doesn't run anything, but it does do all the workspace creation and currently saves the dag_vis outputs into that workspace.  will look into a -no-workspace option

setup.py
FrankD412: I was trying to test this and ended up having to install the requirements manually. Might be worth adding them to the requirements.txt file as pip doesn't seem to allow optional specification when installing from a repository locally.

maestrowf/maestro.py
FrankD412, line 339: It might be worth gating the command line parameter addition behind a package check as seen here: https://stackoverflow.com/a/27361558

FrankD412: Where would the `--no-workspace` option be added. It wouldn't make sense to have that be standalone. It'd have to be explicitly paired with `--dry` and/or `--draw`?

jwhite242: Yeah, i saw that discussion when looking into best practices for this kind of feature disablement.  I'm a bit torn on doing it in function vs setting a file level variable at the top; file level might be more appropriate as I'm thinking the vis function needs splitting (particularly for future use in additional layers on top of maestro where you might want to return/serialize the networkx representation instead of rendering it on the spot).

As for gating the command line parameter, there is a hangup there which led me to doing it in the function: import time.  Doing a bunch of extra import tests in the main driver will most definitely slow things down, whereas when it's gated in the conductor as it is now the user never sees it.  Although it does delay the check and do a lot of work, potentially for nothing by delaying the import checks until right before they're needed..  Will have to dig some more on whether such checks can be done via alternate means to catch it early without always slowing it down.

jwhite242: So, that's a good question.  While that thought/idea hadn't been really fleshed out at the time, I think it'd have to be tied to --draw. Does it really make any sense to have it tied to --dry? That would seem to defeat the purpose of dry runs.

FrankD412: I'm not sure if the import checks really do take that long, might have to test that. But the test upfront let's the user know that the feature isn't enabled. By pushing it to the conductor, the user might not know they don't have it installed and then it silently fails. I think a simple import check for one of the three packages might suffice?

setup.py
jwhite242: Yeah, and graphviz has to be manual no matter what for now, which is kind of annoying; definitely needs some docs for that.  

So, I'd thought the extras provisions in the setup.py would have taken care of that, preventing the need for multiple requirements files..   That's disappointing to hear it doesn't work out that way..

maestrowf/maestro.py
FrankD412: Oh -- good call. I went too far with the logic on that one. I think there is a way to pair CLI paramters, but I'm not sure how that displays in help.

FrankD412: Another suggestion for the command line option in particular -- look at the `choices=[]` optional field for `add_argument`. That lets you enforce specific choices for command line flags.

setup.py
FrankD412: What I meant was that during testing, since I had to install it via `pip install .` it didn't allow me to use the optional flags. I couldn't figure out how to get it to work, and noticed they weren't in the `requirements.txt` -- so those should be there for developers to use. Otherwise, I think it's just compiling a wheel and add that path using `-f` to test is the optionals work I think?

FrankD412:
I tested the `matplotlib` code with the LULESH sample specification and the visualization got messy. I've not tested with the dot format yet, primarily because I need to figure out how to render it. But the image below was the result of my initial run.

dag_lulesh_sample1_

jwhite242:
> I tested the `matplotlib` code with the LULESH sample specification and the visualization got messy. I've not tested with the dot format yet, primarily because I need to figure out how to render it. But the image below was the result of my initial run.
> 
> dag_lulesh_sample1_

Hmm, the autosizing doesn't work so well on that one with those interior nodes.  For the dot format, if you have pygraphviz/graphviz installed you can use `--draw mpl-dot` which will render it for you with graphviz's much nicer dot layout for the tree like structures.     Anyway, I'll take another pass at some more clever sizing (or maybe pull the cleverness out since it looks terrible when it doesn't work).

maestrowf/datastructures/core/study.py
```diff
@@ -164,7 +187,8 @@ class Study(DAG, PickleInterface):
     """
 
     def __init__(self, name, description,
```
FrankD412, line 189: This might be a knit-pick, but it might be better to move `draw=False` to the `configure` method. That's where `throttle` and other settings are set -- to me it makes a bit more sense to have it there.

maestrowf/datastructures/core/study.py
```diff
@@ -164,7 +187,8 @@ class Study(DAG, PickleInterface):
...
-                 studyenv=None, parameters=None, steps=None, out_path="./"):
+                 studyenv=None, parameters=None, steps=None, out_path="./",
+                 draw=False):
```
FrankD412, line 191: This might be a knit-pick, but it might be better to move `draw=False` to the `configure` method. That's where `throttle` and other settings are set -- to me it makes a bit more sense to have it there.

FrankD412: There's also an associated place where we print settings:
```
[2020-08-18 10:29:02: INFO] [study: 460]
------------------------------------------
Submission attempts =       1
Submission restart limit =  1
Submission throttle limit = 0
Use temporary directory =   False
Hash workspaces =           False
Dry run enabled =           True
Output path =               /mnt/f/Code/Python/maestrowf/sample_output/lulesh/lulesh_sample1_20200818-102902
------------------------------------------
```

FrankD412:
> > I tested the `matplotlib` code with the LULESH sample specification and the visualization got messy. I've not tested with the dot format yet, primarily because I need to figure out how to render it. But the image below was the result of my initial run.
> > dag_lulesh_sample1_
> 
> Hmm, the autosizing doesn't work so well on that one with those interior nodes. For the dot format, if you have pygraphviz/graphviz installed you can use `--draw mpl-dot` which will render it for you with graphviz's much nicer dot layout for the tree like structures. Anyway, I'll take another pass at some more clever sizing (or maybe pull the cleverness out since it looks terrible when it doesn't work).

Sounds good -- I ran a rendering in `mpl-dot` and got the following. It's better, but the downstream nodes seem to bundle up.

dag_lulesh_sample1_

jwhite242:
Yeah, the layout algorithm doesn't currently take node text size into account.  That's sort of a pain point currently, but i've got a few ideas for dealing with that that I'm working on.  I think graphviz can deal with that better, and is something i still need to port from my old prototype

maestrowf/datastructures/core/study.py
FrankD412: Alright fixed the above comment:
```
------------------------------------------
Submission attempts =       1
Submission restart limit =  1
Submission throttle limit = 0
Use temporary directory =   False
Hash workspaces =           False
Dry run enabled =           True
DAG drawing enabled =       ['mpl-dot']
Output path =               /mnt/f/Code/Python/maestrowf/sample_output/lulesh/lulesh_sample1_20200818-121610
------------------------------------------
```

maestrowf/maestro.py
FrankD412: Tweaked to have choices and updated help message:
```
  --draw {mpl,mpl-dot,graphml,graphml-dot}
                        Generate visualizations of study's DAG (specify --draw
                        multiple times with various args to draw multiple
                        formats). Supported formats: matplotlib png (mpl), dot
                        layout mpl (mpl-dot), graphviz dot file (dot), graphml
                        file (graphml), dot layout positioned graphml file
                        (graphml-dot)
```

Issue #297, opened 2020-07-14
slurm script adapter does not handle per-step procs correctly


tadesautels:
If I specify a study script like so:

```
batch:
  type: slurm
  queue: pbatch
  bank: abank
  host: amachine

study:
  - name: local-step
    description: locally-executed
    run: 
      cmd: |
          echo "do stuff"
      depends: []
  - name: scheduled-step
     description: scheduled
     run: 
        cmd: |
          echo "Do scheduled stuff"
        procs: 4
        walltime: 24:00:00
        depends: [local-step]
```

The SLURM scheduled step is scheduled, but the job does not have a `# SBATCH --ntasks=4`  line specifying the number of tasks.

However, if i add `procs: 4` to the batch line, this step is scheduled (the previous local one is correctly NOT scheduled) and the  `# SBATCH -n 4` appears correctly.

FrankD412:
The behavior you described is expected in regards to `procs`. If `procs` is not present in the `batch` section, Maestro isn't "hinted" to include it in the headers of the scripts. Generally, users take full nodes and use srun to control the resource usage from what I've seen. What's interesting is that the specification of `procs` in `batch` seems to control the presence of `nodes` which gives me an idea of where this is happening.

FrankD412:
@tadesautels -- I was look through tickets, and noticed; you've not specified nodes in either the `batch` section or the `step` that you intend to schedule on nodes. In SLURM's case, it's not possible to continue -- is it fair to say that the adapter should throw an error in this case?

tadesautels:
Agreed, it's probably best to force the maestro study to specify nodes _somewhere_ for SLURM.
My only hesitation is that in this particular case, I was trying to work on a machine where you do not reserve whole nodes, but instead use the ntasks (-n) flag to specify the number of processes; -N might even be ignored on this machine (I can't tell from the machine-specific documentation).  It's a bit of an peculiar case. 

Unfortunately, while we did some debugging on that machine, I can't seem to find the maestro studies that actually ran through correctly.

PR #305, opened 2020-08-20
Slurm Script Adapter Bug Fix

youngjeffrey:
https://github.com/LLNL/maestrowf/issues/297


maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -71,22 +72,38 @@ def __init__(self, **kwargs):
         """
         super(SlurmScriptAdapter, self).__init__(**kwargs)
 
+        # If Nodes is greater than 0, add it to the header
+        nodes = kwargs.pop("nodes", "0")
+        if nodes !- "0":
```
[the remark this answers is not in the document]
FrankD412: Tom's comment [here](https://github.com/LLNL/maestrowf/issues/297#issuecomment-673098513):
> Agreed, it's probably best to force the maestro study to specify nodes _somewhere_ for SLURM.
> My only hesitation is that in this particular case, I was trying to work on a machine where you do not reserve whole nodes, but instead use the ntasks (-n) flag to specify the number of processes; -N might even be ignored on this machine (I can't tell from the machine-specific documentation). It's a bit of an peculiar case.
> 
> Unfortunately, while we did some debugging on that machine, I can't seem to find the maestro studies that actually ran through correctly.

So, I think this has to be modified in two ways:
1. We shouldn't conditionally add this to the header, it should always be present. If `nodes` is not specified, in theory it's either 0 or an empty string. This check in `get_header` will pass it if it isn't specified so it's safe to have here.
2. This fulfills the comment linked above where `nodes` will not be present if not specified, meaning a user can specify to not use node allocation on machines that do not require it.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -149,6 +136,17 @@ def get_header(self, step):
         if exclusive:
             modified_header.append("#SBATCH --exclusive")
 
+        # If neither Procs nor Nodes exist, throw an error
+        procs = resources.get("procs")
+        nodes = resources.get("nodes")
+        if procs == None and nodes == None:
```
FrankD412, line 142: This condition won't work because `procs` and `nodes` are never set to `None`. You also can't achieve that with the `get` call because they will exist in the resources as empty strings (`""`). Try this conditional: `if not procs and not nodes:` -- this clause will work because Python equates `""` to a "False" value in an `if` clause.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -149,6 +136,17 @@ def get_header(self, step):
...
+            err_msg = "At least one of either Procs or Nodes needs to be "
+            "present for Script to proceed."
+            LOGGER.error(err_msg)
+            rt_err_msg = "No explicit resources specified. At least one of " 
+            "Nodes or Procs must be set to a non-zero value."
```
FrankD412, line 147: The message is fine but should be made clearer actually. When we talked earlier, I forgot to mention that we can use the step's entries to write a specific error message. In this case, make sure that nodes and procs are capitalized exactly as they are in the specification. You can also use the step to print the name of the step that's missing these entries so that the user knows exactly where to look.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -128,15 +132,21 @@ def get_header(self, step):
...
 
+        # If neither Procs nor Nodes exist, throw an error
+        procs = resources.get("procs")
+        nodes = resources.get("nodes")
+        if not procs and not nodes:
+            err_msg = "At least one of either (procs) or (nodes) in {} needs "
+            "to be present for Script to proceed.".format(step.name)
+            LOGGER.error(err_msg)
+            rt_err_msg = "No explicit resources specified in {}. At least one of " 
+            "(nodes) or (procs) must be set to a non-zero value.".format(step.name)
```
FrankD412, line 147: Sorry, I entirely mis-typed. When I said parentheses I totally meant quotations. Sorry about that. You can also use the same error message for both the logging and exception.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -140,10 +140,14 @@ def get_header(self, step):
...
         if not procs and not nodes:
-            err_msg = "No explicit resources specified in {}. At least one of " 
-            "\"procs\" or \"nodes\" must be set to a non-zero value.".format(step.name)
+            err_msg = \
+                'No explicit resources specified in {}. At least one' \
+                ' of "procs" or "nodes" must be set to a non-zero' \
+                ' value.'.format(step.name)
             LOGGER.error(err_msg)
             raise RuntimeError(err_msg)
+        elif procs or (procs and not nodes):
```
FrankD412, line 149: This conditional is slightly incorrect -- you need to check if `procs` is in the `batch` section, not if it's non-zero.

youngjeffrey: You left a comment "This conditional is slightly incorrect -- you need to check if procs is in the batch section, not if it's non-zero."
In regards to this comment...

From my notes, the goal of this conditional check is because SBATCH ntasks is not present even when 'procs: 4' is in the step. So if procs is in the step, doesn't that mean it won't be in the batch section. In addition, procs is obtained via 'resources.get('procs') and resources is composed of batch and run dicts.

FrankD412: > From my notes, the goal of this conditional check is because SBATCH ntasks is not present even when 'procs: 4' is in the step. So if procs is in the step, doesn't that mean it won't be in the batch section. In addition, procs is obtained via 'resources.get('procs') and resources is composed of batch and run dicts.

The `procs` key can be in both the batch and the step. So the only time that `procs` part of the conditional will be false is if it's not set in either.

So I noticed that in the `__init__` you also check for the `procs` in the `batch` section. With this conditional here, you will add `procs` to the header twice. I don't think SLURM will throw an error, but the header should be as clean as possible.

In `__init__` instead of adding the `--ntasks` key, add it here. Simply save the `--ntasks` header line in the `__init__` so that it's in the same place as everything else. Something like:
```
def __init__(self, **kwargs):
    ...
    ntask_header = "#SBATCH --ntasks={procs}"
    ...
```

A suggestion on the exception: move to be right after the chainmap construction -- that'll make it so the header throws the exception without the extra logic. If that exception case is try, we can't do anything. 

After we do the loop for adding items to the header, `nodes` will be there if it's specified (it's set to `""` by default) otherwise we won't have `procs` in the header at all at this point. This is now where you should check for both the hint (`procs in self._batch`) or if `nodes` is not set (because we need to give SLURM some sort of resource to use). In theory, you can even move the `exclusive` addition to be later as well so that you're handling resources in one block.

Does that make sense?

FrankD412:
There's a subtle bug here too -- it looks like the `nodes` isn't being added to the batch settings when the adapter is stood up. Might be worth verifying that works.

youngjeffrey:
In regards to `nodes `not being added to the batch settings, I think this is because `self.add_batch_parameters("nodes", kwargs.pop("nodes", 1))` got removed. What does it mean when "adapter is stood up"

FrankD412:
@youngjeffrey -- Sorry, but stood up I mean when the `SlurmScriptAdapter` is instantiated.

maestrowf/interfaces/script/slurmscriptadapter.py
youngjeffrey: With the way that the code is structured right now, if procs is not specified in the `batch.info`, it will never be in the `self._batch`. This means that it cannot be added to the header because the condition is not met.

youngjeffrey:
In regards to my latest commit, `nodes `has been added again to `self._batch `if it is present or as an empty string `""` if it is not. It is now in the batch settings. It also helps with the exception check later because empty string equates to if not nodes:. Checking if `procs `is in` self._batch` has been moved once again to `get_header`, this time with the additional condition `or not nodes`. I did a sample test in ipython and it seems to work where the `procs `is added to the header.

FrankD412:
Sorry it took me so long to get back, have been busy with other things. I found a case that doesn't work (see below). It turns out that the use of the `Chainmap` bites us here. Because the `chainmap` doesn't care if a value is empty, a step that comes in with an uninitialized `nodes` key still has the `""` in the field, meaning that the key exists. That then means if `nodes` is specified in the `batch` section it gets hidden behind the empty string. It looks like we might have to do away with the `chainmap` and revert to static checking.

```
description:
    name: slurm_test
    description: Sample test for SLURM header

batch:
    type: slurm
    queue: pbatch
    bank: abank
    host: amachine
    nodes: 1
  
study:
    - name: local-step
      description: locally-executed
      run: 
          cmd: |
            echo "do stuff"
          depends: []
    - name: scheduled-step
      description: scheduled
      run: 
          cmd: |
            echo "Do scheduled stuff"
          walltime: "24:00:00"
          procs: 4
          depends: [local-step]
```

youngjeffrey:
> Sorry it took me so long to get back, have been busy with other things. I found a case that doesn't work (see below). It turns out that the use of the `Chainmap` bites us here. Because the `chainmap` doesn't care if a value is empty, a step that comes in with an uninitialized `nodes` key still has the `""` in the field, meaning that the key exists. That then means if `nodes` is specified in the `batch` section it gets hidden behind the empty string. It looks like we might have to do away with the `chainmap` and revert to static checking.
> 
> ```
> description:
>     name: slurm_test
>     description: Sample test for SLURM header
> 
> batch:
>     type: slurm
>     queue: pbatch
>     bank: abank
>     host: amachine
>     nodes: 1
>   
> study:
>     - name: local-step
>       description: locally-executed
>       run: 
>           cmd: |
>             echo "do stuff"
>           depends: []
>     - name: scheduled-step
>       description: scheduled
>       run: 
>           cmd: |
>             echo "Do scheduled stuff"
>           walltime: "24:00:00"
>           procs: 4
>           depends: [local-step]
> ```

Do you think it's better to replace the `chainmap` altogether or add an extra conditional statement about `nodes `in the for loop where I add resources to the `modified_header`

youngjeffrey:
Nodes now show up on the` SBATCH header` if they are initialized in `Batch `and not initialized in the `Step`. The `chain map `remains, I added a for loop to check for keys which had a value of `""`. Then I checked which key in the `Batch Map` or `Step Map `had the empty string or was set to -1 (done to avoid KeyError). I would then update the other key to be in the `resources `map. 

Later in the code, I slightly changed the conditional statement for adding `procs `to the `modified header`.  This is because `ntasks `would no longer show up if `procs `was not initialized in Batch but was initialized in Step.

FrankD412:
> Do you think it's better to replace the `chainmap` altogether or add an extra conditional statement about `nodes `in the for loop where I add resources to the `modified_header`

Honestly, with how the `StudyStep` is set up it's probably not the use case for the `chainmap` at this point. I had used it because it seemed like it fit, but now I realize that it actually didn't serve the use case at all. It's probably cleaner to get rid of it at this point because the other alternative is to override the `__getattribute__` magic method on the `StudyStep` which can get hairy if you're not careful.

FrankD412:
We should also rebase your branch so that you get the latest tests, so that you can actually run the tests.

youngjeffrey:
`Resources `is now a dictionary instead of a chain map. `SBATCH header` still remains the same and solves the previous issue of node not appearing in the batch.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -116,22 +117,49 @@ def get_header(self, step):
         :returns: A string of the header based on internal batch parameters and
             the parameter step.
         """
-        resources = ChainMap(step.run, self._batch)
+
+        resources = {}
+        resources.update(step.run)
```
FrankD412, line 122: This produces incorrect behavior because the update is backwards here -- in this order, the `batch` settings will override the more accurate and specific step settings. I just tested it, and this is in fact the case. I would do something like:

```
resources = {}
resources.update(self._batch)
resources.update({ key : value for key, value in step.run if value })
```

From there, you can check `procs` and `nodes` and throw the exception if both aren't set, run through the resources, and then add `ntasks` if needed.

FrankD412: Sorry, I think I got the dictionary comprehension wrong last night -- it should look something like:
```
resources.update({ key : value for (key, value) in step.run if value })
```

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -32,6 +32,7 @@
 import logging
 import os
 import re
+import sys
 # In order to support Python2.7, we need to catch the import error.
 try:
     from collections import ChainMap
```
FrankD412, line 38: This should be removed, it's no longer used.

FrankD412:
@youngjeffrey -- I just made some commits to change the construction of the resources and capture if `procs` is in the batch. Take a look, but my initial testing shows that it works.

```
description:
  name: slurm_test
  description: Sample test for SLURM header

batch:
  type: slurm
  queue: pbatch
  bank: abank
  host: amachine
  #nodes: 1
  procs: 1

study:
  - name: local-step
    description: locally-executed
    run: 
        cmd: |
          echo "do stuff"
        depends: []
  - name: scheduled-step
    description: scheduled
    run: 
        cmd: |
          echo "Do scheduled stuff"
        walltime: "24:00:00"
        procs: 4
        nodes: 2
        depends: [local-step]
  - name: scheduled-step_2
    description: scheduled
    run: 
        cmd: |
          echo "Do scheduled stuff"
        walltime: "24:00:00"
        procs: 5
        depends: [local-step]
```

PR #306, opened 2020-08-21
Updated the documentation to pull the recent version

jsemler:
Updated the Sphinx documentation to pull Maestro’s version from the package.


docs/source/conf.py
```diff
@@ -19,6 +19,8 @@
 import sys
 sys.path.insert(0, os.path.abspath('../../'))
 
+from maestrowf import __version__, __version_info__
```
FrankD412, line 22: It looks likes this line needs to be moved up past the first `sys.path` call -- `flake8` doesn't like where this is.

jsemler: Fixed.

PR #314, opened 2020-09-11
Add name and real_name StudyStep properties.

### LINKED ISSUE #267: when using hashws, shouldn't the script names (e.g. slurm adapter) be hashed also?
i have a use case with lots of variables for variation. we had to move to the hashed workspace option because the paths were too long.  i've run into a failure when using the hashed ws, that the script creation fails because it violates the 256 character limitation of file name length!  
if the dirs are hashed, shouldn't/couldn't the script names be hashed?  
i believe you already have a correspondence setup between steps names, workspaces and hash.  just use the hash for the script name.
Thanks!

### LINKED ISSUE #291: Unicode hashing error
  File "3.6-maestro_dev/bin/maestro", line 11, in <module>
    load_entry_point('maestrowf', 'console_scripts', 'maestro')()
  File "tools/maestrowf_dev/maestrowf/maestro.py", line 430, in main
    rc = args.func(args)
  File "tools/maestrowf_dev/maestrowf/maestro.py", line 274, in run_study
    conductor.initialize(batch, sleeptime)
  File "tools/maestrowf_dev/maestrowf/conductor.py", line 303, in initialize
    self._pkl_path, self._exec_dag = self._study.stage()
  File "tools/maestrowf_dev/maestrowf/datastructures/core/study.py", line 853, in stage
    return self._out_path, self._stage(dag)
  File "tools/maestrowf_dev/maestrowf/datastructures/core/study.py", line 634, in _stage
    *[step, md5(combo_str).hexdigest()])
TypeError: Unicode-objects must be encoded before hashing

FrankD412:
There now needs to be an aliasing of nickname with a step name when requested because adapters expect to use the name. StudyStep objects now return their nickname if the nickname is set. Objects such as the ExecutionGraph must now use the real_name for logistic tracking, while adapters can continue to use name. The nickname field is then used to fill in hashing nicknames to correct the script generation.

Fixes #291
Fixes #267

### DIFF
--- a/maestrowf/abstracts/interfaces/scriptadapter.py
+++ b/maestrowf/abstracts/interfaces/scriptadapter.py
@@ -28,7 +28,7 @@
 ###############################################################################
 
 """Abstract Script Interfaces for generating scripts."""
-from abc import ABCMeta, abstractmethod
+from abc import ABCMeta, abstractmethod, abstractproperty
 import logging
 import os
 import six
@@ -121,6 +121,14 @@ class ScriptAdapter(object):
             st = os.stat(restart_path)
             os.chmod(restart_path, st.st_mode | stat.S_IXUSR)
 
+        LOGGER.debug(
+            "---------------------------------\n"
+            "Script path:   %s"
+            "Restart path:  %s"
+            "Scheduled?:    %s"
+            "---------------------------------\n",
+            script_path, restart_path, to_be_scheduled
+        )
         return to_be_scheduled, script_path, restart_path
 
     @abstractmethod
@@ -143,8 +151,16 @@ class ScriptAdapter(object):
         """
         pass
 
-    @property
-    @abstractmethod
+    @abstractproperty
+    def extension(self):
+        """
+        Returns the extension that generated scripts will use.
+
+        :returns: A string of the extension
+        """
+        pass
+
+    @abstractproperty
     def key(self):
         """
         Return the key name for a ScriptAdapter..
--- a/maestrowf/abstracts/interfaces/scriptadapter.py
+++ b/maestrowf/abstracts/interfaces/scriptadapter.py
@@ -123,9 +123,9 @@ class ScriptAdapter(object):
 
         LOGGER.debug(
             "---------------------------------\n"
-            "Script path:   %s"
-            "Restart path:  %s"
-            "Scheduled?:    %s"
+            "Script path:   %s\n"
+            "Restart path:  %s\n"
+            "Scheduled?:    %s\n"
             "---------------------------------\n",
             script_path, restart_path, to_be_scheduled
         )
--- a/maestrowf/datastructures/core/study.py
+++ b/maestrowf/datastructures/core/study.py
@@ -67,8 +67,9 @@ class StudyStep:
 
     def __init__(self):
         """Object that represents a single workflow step."""
-        self.name = ""
+        self._name = ""
         self.description = ""
+        self.nickname = ""
         self.run = {
                         "cmd":              "",
                         "depends":          "",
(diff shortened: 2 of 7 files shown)


maestrowf/datastructures/core/study.py
```diff
@@ -97,6 +98,35 @@ def apply_parameters(self, combo):
...
+        if self.nickname:
+            return self.nickname
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        """
+        Set the name of a StudyStep instance.
+
+        :oaram value: A string value representing the name to give the step.
```
jwhite242, line 117: typo: oaram->param

maestrowf/datastructures/core/study.py
```diff
@@ -335,7 +365,7 @@ def add_step(self, step):
         :param step: A StudyStep instance to be added to the Study instance.
         """
         # Add the node to the DAG.
-        self.add_node(step.name, step)
+        self.add_node(step.real_name, step)
```
jwhite242, line 368: why not just use step.name?  won't nickname need to be unique as well for it to work?   Or is it just for easier debugging since hashes aren't the most readable things?

maestrowf/datastructures/core/study.py
```diff
@@ -625,11 +655,15 @@ def _stage(self, dag):
                                 "**********************************",
                                 str(combo))
                     # Compute this step's combination name and workspace.
+                    nickname = None
                     combo_str = combo.get_param_string(self.used_params[step])
+                    # We must encode explicitly to utf-8
+                    # combo_str = combo_str.encode("utf-8")
                     if self._hash_ws:
```
jwhite242, line 662: Can this be made to turn on automatically if names go beyond a certain length?  might be a nice safety valve for the users so they don't get a bunch of paths that are too long without realizing it?  Or maybe a better check is on the filename lenghts of the shell scripts since those are much shorter than the full path limits: looks like ~250 on linux, path is 4096, but other os's appear to be lower on that.

maestrowf/abstracts/interfaces/scriptadapter.py
```diff
@@ -121,6 +121,14 @@ def write_script(self, ws_path, step):
             st = os.stat(restart_path)
             os.chmod(restart_path, st.st_mode | stat.S_IXUSR)
 
+        LOGGER.debug(
+            "---------------------------------\n"
+            "Script path:   %s"
```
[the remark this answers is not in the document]
FrankD412: Good call -- will add those here in a second.

maestrowf/datastructures/core/study.py
FrankD412: That's true -- and I had considered using the hash. It's a mix of wanting to keep things behaving as close to what they were before and the debugging.

maestrowf/datastructures/core/study.py
FrankD412: That's an interesting consideration. My main concern with that is some users might use `grep` to find parameter folders and things in post processing. I also don't know a good way to absolutely make sure a user knows that they are being forced into hashed workspaces. This might be worth opening an ideas discussion about and tagging folks who might have something to say on this.

jwhite242: well, on the flip side, if the length is too long grepping won't work anyway.   but, not being aggressive with the threshold would be nice as no-hashed makes things easier to navigate with visit and the like too.

definitely worth discussing the how: most of this would happen in background currently right?  makes debug logging as the warning mechanism a bit troublesome

FrankD412: Yeah, exactly. The only way that I can think would be to warn the user in the Maestro front end when the number of parameters is large and has a risk of going over. Maybe then ask them a yes/no if they want to hash workspaces. Mind starting an ideas discussion?

jwhite242 (approved):

maestrowf/datastructures/core/study.py
jwhite242: Sure; I just created a discussion so we can keep going after the PR

PR #315, opened 2020-09-11
Removal of Python 2.7 and addition of newer Python 3 versions

FrankD412:
Maestro will be dropping support for Python 2 in the upcoming release. This PR updates the tests since the `setup.py` file has now been updated and it's causing TravisCI failures in PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
 
 python:
-    - "2.7"
     - "3.5"
     - "3.6"
+    - "3.7"
+    - "3.8"
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python
 python:
     - "3.5"
     - "3.6"
-    - "3.7"
     - "3.8"
 
 matrix:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: python
 
 python:
-    - "2.7"
     - "3.5"
     - "3.6"
+    - "3.8"
```
kcathey, line 6: Don't you also need to add 3.7 here as well?

kcathey (changes requested):

.travis.yml
FrankD412: So when I added it, Travis actually duplicated it. I'm not 100% sure why.

FrankD412: Actually let me remove the custom setting from the files and just rely on Travis to do it...

kcathey: I think that is because of the matrix include below.

FrankD412: Yeah, that's exactly what I'm thinking. That was an old workaround for Travis not launching 3.7 -- might be fixed now.

FrankD412:
@kcathey -- Yeah looks like Travis fixed that old issue. Fixed.

kcathey:
> @kcathey -- Yeah looks like Travis fixed that old issue. Fixed.

Awesome. Glad they publicized that.

kcathey (approved):

Issue #310, opened 2020-08-25
Add the job identifier to the status print out


FrankD412:
A closely related request would be to include job ids in "maestro status"

_Originally posted by @crkrenn in https://github.com/LLNL/maestrowf/issues/304#issuecomment-676865385_

FrankD412:
Closed by #323

PR #323, opened 2020-09-30
WIP: Add job identifier to status command

youngjeffrey:
https://github.com/LLNL/maestrowf/issues/310

### DIFF
--- a/maestrowf/conductor.py
+++ b/maestrowf/conductor.py
@@ -227,7 +227,6 @@ class Conductor:
         study_glob = \
             glob.glob(os.path.join(out_path, "*{}".format(cls._pkl_extension)))
 
-        print(study_glob)
         if len(study_glob) == 1:
             # We only expect one result.If we only get one, let's assume and
             # check after.
@@ -356,7 +355,8 @@ class Conductor:
             dag.write_status(os.path.split(pkl_path)[0])
             # Sleep for SLEEPTIME in args if study not complete.
             if completion_status == StudyStatus.RUNNING:
-                sleep(sleep_time)
+                #sleep(sleep_time)
+                sleep(3)
 
         return completion_status
 
--- a/maestrowf/conductor.py
+++ b/maestrowf/conductor.py
@@ -355,8 +355,7 @@ class Conductor:
             dag.write_status(os.path.split(pkl_path)[0])
             # Sleep for SLEEPTIME in args if study not complete.
             if completion_status == StudyStatus.RUNNING:
-                #sleep(sleep_time)
-                sleep(3)
+                sleep(sleep_time)
 
         return completion_status
 
--- a/maestrowf/datastructures/core/executiongraph.py
+++ b/maestrowf/datastructures/core/executiongraph.py
@@ -602,16 +602,21 @@ class ExecutionGraph(DAG, PickleInterface):
     def write_status(self, path):
         """Write the status of the DAG to a CSV file."""
         header = "Step Name,Workspace,State,Run Time,Elapsed Time,Start Time" \
-                 ",Submit Time,End Time,Number Restarts"
+                 ",Submit Time,End Time,Number Restarts, Job ID"
         status = [header]
         keys = set(self.values.keys()) - set(["_source"])
         for key in keys:
             value = self.values[key]
+
+            jobid_str = "Job ID not available"
+            if value.jobid:
+                jobid_str = " ".join(value.jobid)
+
             _ = [
                     value.name, os.path.split(value.workspace.value)[1],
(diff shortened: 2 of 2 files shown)


maestrowf/conductor.py
```diff
@@ -356,7 +355,8 @@ def monitor_study(self):
             dag.write_status(os.path.split(pkl_path)[0])
             # Sleep for SLEEPTIME in args if study not complete.
             if completion_status == StudyStatus.RUNNING:
-                sleep(sleep_time)
+                #sleep(sleep_time)
+                sleep(3)
```
FrankD412, line 359: You need to change back to the old sleep -- this breaks a feature in Maestro.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -602,16 +602,21 @@ def _execute_record(self, record, adapter, restart=False):
     def write_status(self, path):
         """Write the status of the DAG to a CSV file."""
         header = "Step Name,Workspace,State,Run Time,Elapsed Time,Start Time" \
-                 ",Submit Time,End Time,Number Restarts"
+                 ",Submit Time,End Time,Number Restarts, Job ID"
```
FrankD412, line 605: I feel as though the `JobID` should be after the `Step Name` since they're both identifiers. Thoughts?

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -602,16 +602,21 @@ def _execute_record(self, record, adapter, restart=False):
...
         status = [header]
         keys = set(self.values.keys()) - set(["_source"])
         for key in keys:
             value = self.values[key]
+
+            jobid_str = "Job ID not available"
+            if value.jobid:
+                jobid_str = " ".join(value.jobid)
```
FrankD412, line 613: A couple things here:
1. `Job ID not available` is a little long for a column. Identifiers are under 10 digits long. That's way over 10. It's probably better to leave the column blank, or have something like `--` (I think the steps that haven't run have something like `--:--:--` for their timestamps). You don't need to be overly verbose, a visual cue is enough to tell a user isn't not available.
2. We should only report the last one. If something is running and has run multiple times, the user is likely to not care about the old identifiers. The most recent one is the most important. In that case, simply report `value.jobid[-1]` if the list is non-empty.

FrankD412 (changes requested):
You should also update your fork from `upstream`. It looks like you have a branch based off of an older fork, so you may want to rebase on `origin/develop` to get the latest. I'll have to help you get set up with some of the recent changes.

maestrowf/datastructures/core/executiongraph.py
FrankD412, line 613: As an FYI, when joining a list -- the items must be strings. It's an easy gotcha in Python.

Your current code fails here:
```
  File "/mnt/f/Code/Python/maestrowf/maestrowf/conductor.py", line 382, in main
    completion_status = conductor.monitor_study()
  File "/mnt/f/Code/Python/maestrowf/maestrowf/conductor.py", line 355, in monitor_study
    dag.write_status(os.path.split(pkl_path)[0])
  File "/mnt/f/Code/Python/maestrowf/maestrowf/datastructures/core/executiongraph.py", line 613, in write_status
    jobid_str = " ".join(value.jobid)
TypeError: sequence item 0: expected str instance, int found
[2020-09-30 23:58:04,209: INFO] Study exiting, cleaning up...
[2020-09-30 23:58:04,209: INFO] Squeaky clean!
```

maestrowf/conductor.py
youngjeffrey: Oh yeah sorry, I changed that to do some iPython testing. I'll change it back

maestrowf/datastructures/core/executiongraph.py
youngjeffrey: I think that's more appropriate too

maestrowf/datastructures/core/executiongraph.py
youngjeffrey: Sounds good, I will make that change

FrankD412:
@youngjeffrey -- I fixed the casting issue with the PR. It appears to be working.

FrankD412 (approved):

crkrenn:
@youngjeffrey & @FrankD412 

Hello Jeffrey and Frank, 

Thanks very much for adding the job id. I'm looking forward to using this!

I have a related request: Can you also add machine to the list of information given by status? I'm running on 3 different machines on the SCF, and it would be very helpful to know which jobs are running on which machines. 

Thanks!

-Chris

PR #331, opened 2020-10-30
Global parameter reconstruction utility.

vashdv:
Adds the ability to generate a global parameters block given a study meta/parameters.yaml.

### DIFF
--- a/maestrowf/utils.py
+++ b/maestrowf/utils.py
@@ -402,7 +402,7 @@ class ReconstructGlobalParameters:
 
         return (self.formatted_names, self.formatted_values, self.formatted_labels)
 
-    def build_parameters(self, formatted_items):
+    def build_parameters(self, formatted_items: tuple(list, list, list)):
         """Builds a global.parameters block given formatted names, values, and labels then saves to generated_global_parameters.yaml.
 
         Example:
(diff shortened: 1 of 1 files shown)


maestrowf/utils.py
```diff
@@ -349,3 +350,95 @@ def map_level(log_lvl):
             return logging.ERROR
         else:
             return logging.CRITICAL
+
+
+class ReconstructGlobalParameters:
```
FrankD412, line 355: It may be worth renaming this to something more general like `MetadataParser` -- this class could be used as the starting point for other metadata parsing which contains methods for global parameter parsing. But I can easily envision this being extended to load in the metadata and search for workspaces, for example.

maestrowf/utils.py
```diff
@@ -349,3 +350,95 @@ def map_level(log_lvl):
...
+        values: list = formatted_items[1]
+        labels: list = formatted_items[2]
+
+        for i in range(len(values)):
+            blocks: dict = {
+                'name': names[i],
+                'value': values[i],
+                'label': labels[i]
+            }
+            formatted_block: str = \
```
FrankD412, line 431: I think that we discussed this in a 1:1 a bit ago before I had to take some leave, but it looks like you've manually formatted a YAML line here. It might be worth compiling this into a directory structure that matches the specification, then use a markup language specific library to dump it to its appropriate format.

FrankD412:
@vashdv -- I'm closing this PR since it's been inactive for a year. Feel free to open it back up if you pick this back up in the future.

PR #338, opened 2021-01-30
Addition of cancellation for multiple directories and globs.

FrankD412:
The `maestro cancel` subcommand now accepts globbing and multiple paths for cancellation. It simply places a cancel lock in each directory just as it did for a single directory previously. Additionally, it will ask the user if they are sure they would like to kill the printed directories before actually canceling.

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -103,16 +103,31 @@ def cancel_study(args):
     directory_list = args.directory
 
     ret_code = 0
+    to_cancel = []
     if directory_list:
         for directory in directory_list:
-            if not os.path.isdir(directory):
+            abs_path = os.path.abspath(directory)
+            if not os.path.isdir(abs_path):
                 print(
-                    f"Attempted to cancel '{directory}' "
+                    f"Attempted to cancel '{abs_path}' "
                     "-- study directory not found.")
-                ret_code = 2
+                ret_code = 1
             else:
-                print(f"Marked study in '{directory}' to be cancelled.")
-                Conductor.mark_cancelled(directory)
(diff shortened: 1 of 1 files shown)


kcathey (commented):
From what I can tell it looks good. Are there tests that go with this?

FrankD412:
I tested this by launching multiple studies and cancelling them all at once. To test the actual functionality automatically, not currently. I'm not sure we would want to test this particular thing in `pytest` -- any thoughts?

kcathey:
I would think the big thing to test with `pytest` would be `Conductor.mark_cancelled()` as that is the actual unit doing the work. The loop here is fairly straight forward so that shouldn't be a big deal.

FrankD412:
I was thinking the same -- alright, I'll go ahead and merge this in. I need to add a testing project, it's getting to be that time.

kcathey (approved):

jwhite242:
so how does the cancel glob work for sub-selecting things to cancel?  also, how about an entry on the docs for using this with the globs, perhaps in the quick start guide example on [cancelling a study](https://maestrowf.readthedocs.io/en/latest/quick_start.html?highlight=cancel#cancelling-a-running-study)

FrankD412:
This gets resolved beforehand by the shell actually -- something we discovered with globs for the status command. You can get as specific using the globbing as you'd like.

That's a good point. I'll have to go back and add it.

jwhite242:
also and idea for a potential feature here (probably a actual keyword arg) could be to enable filtering on parameter values? unsure if that's actually something that'd get used a lot though.

PR #345, opened 2021-02-23
Updates to use pyaestro to back a local pool style adapter.

FrankD412:
This PR is a preliminary step towards a more "ensemble"-ish adapter related to the functionality discussed in #330.

This new `LocalPoolAdapter` utilizes pyaestro's `Executor` class in order to fake a scheduler. Some potential future additions are timeouts and resource support for the above #330.

### DIFF
--- a/maestrowf/interfaces/script/localpooladapter.py
+++ b/maestrowf/interfaces/script/localpooladapter.py
@@ -1,0 +1,204 @@
+import logging
+import os
+
+from maestrowf.abstracts.interfaces import SchedulerScriptAdapter
+from maestrowf.abstracts.enums import JobStatusCode, State, SubmissionCode, \
+    CancelCode
+from maestrowf.interfaces.script import CancellationRecord, SubmissionRecord
+from pyaestro.utilities.executor import Executor, ExecTaskState, ExecCancel
+
+LOGGER = logging.getLogger(__name__)
+
+
+class LocalPoolAdapter(SchedulerScriptAdapter):
+    """Interface class for the flux scheduler (on Spectrum MPI)."""
+
+    key = "local_pool"
+
+    def __init__(self, **kwargs):
+        """
+        Initialize an instance of the LocalPool Adapter.
+
+        The expected keyword arguments that are expected when the Flux adapter
+        is instantiated are as follows:
+        * num_processes: The number of local workers for processing.
+
+        :param **kwargs: A dictionary with default settings for the adapter.
+        """
+        super(LocalPoolAdapter, self).__init__(**kwargs)
+        num_workers = kwargs.get("num_workers", "1")
+        self.add_batch_parameter("max_workers", num_workers)
+        self._pool = Executor(num_workers)
+        self._extension = ".lp.sh"
+
+    @property
+    def extension(self):
+        return self._extension
+
+    def get_header(self, step):
+        """
+        Generate the header present at the top of Flux execution scripts.
+
+        :param step: A StudyStep instance.
+        :returns: A string of the header based on internal batch parameters and
+                  the parameter step.
+        """
+        return f"#!{self._exec}"
+
+    def get_parallelize_command(self, procs, nodes=None, **kwargs):
+        """
+        Generate an empty parallelize command for local execution.
+
+        :param procs: Number of processors to allocate to the parallel call.
+        :param nodes: Number of nodes to allocate to the parallel call
+                      (default = 1).
+        :returns: A string of the parallelize command configured using nodes
+                  and procs.
+        """
+        return ""
+
+    def submit(self, step, path, cwd, job_map=None, env=None):
+        """
+        Submit a script to the Flux scheduler.
+
+        :param step: The StudyStep instance this submission is based on.
+        :param path: Local path to the script to be executed.
+        :param cwd: Path to the current working directory.
+        :param job_map: A dictionary mapping step names to their job
+                        identifiers.
+        :param env: A dict containing a modified environment for execution.
+        :returns: The return status of the submission command and job
+                  identifier.
+        """
+        try:
+            job_id = self._pool.submit(path, cwd)
+            ret_code = 0
+            submit_status = SubmissionCode.OK
+        except Exception as exception:
+            LOGGER.error("Encountered exception: %s", str(exception))
+            job_id = -1
+            ret_code = -1
+            submit_status = SubmissionCode.ERROR
+
+        return SubmissionRecord(submit_status, ret_code, job_id)
+
+    def check_jobs(self, joblist):
+        """
+        For the given job list, query execution status.
+
+        This method uses the scontrol show job <jobid> command and does a
+        regex search for job information.
+
+        :param joblist: A list of job identifiers to be queried.
+        :returns: The return code of the status query, and a dictionary of job
+                  identifiers to their status.
+        """
+        LOGGER.debug("Joblist type -- %s", type(joblist))
+        LOGGER.debug("Joblist contents -- %s", joblist)
+        if not joblist:
+            LOGGER.debug("Empty job list specified.")
+            return JobStatusCode.OK, {}
+        if not isinstance(joblist, list):
+            LOGGER.debug("Specified parameter is not a list.")
+            if isinstance(joblist, int):
+                LOGGER.debug("Integer found.")
+                joblist = [joblist]
+            else:
+                LOGGER.debug("Unknown type. Returning an error.")
+                return JobStatusCode.ERROR, {}
+
+        status = {jid: State.UNKNOWN for jid in joblist}
+        try:
+            for job_id, state in self._pool.get_all_status():
+                if job_id in status:
+                    status[job_id] = self._state(state)
+            chk_status = JobStatusCode.OK
+
+        except Exception as exception:
+            LOGGER.error(str(exception))
+            chk_status = JobStatusCode.ERROR
+
+        return chk_status, status
+
+    def cancel_jobs(self, joblist):
+        """
+        For the given job list, cancel each job.
+
+        :param joblist: A list of job identifiers to be cancelled.
+        :returns: The return code to indicate if jobs were cancelled.
+        """
+        # If we don"t have any jobs to check, just return status OK.
+        if not joblist:
+            return CancelCode.OK
+
+        c_status = set()
+        for job in joblist:
+            c_status.add(self._pool.cancel(job))
+
+        c_return = CancelCode.OK
+        ret_code = 0
+        if ExecCancel.FAILED in c_status:
+            c_return = CancelCode.ERROR
+            ret_code = -1
+        elif ExecCancel.JOBNOTFOUND in c_status:
+            ret_code = 1
+
+        return CancellationRecord(c_return, ret_code)
+
+    def _state(self, executor_state):
+        """
+        Map a scheduler specific job state to a Study.State enum.
+
+        :param executor_state: Enum representation of scheduler job status.
+        :returns: A Study.State enum corresponding to parameter job_state.
+        """
+        if executor_state == ExecTaskState.PENDING:
+            return State.PENDING
+        elif executor_state == ExecTaskState.RUNNING:
+            return State.RUNNING
+        elif executor_state == ExecTaskState.INITIALIZED:
+            return State.INITIALIZED
+        elif executor_state == ExecTaskState.CANCELLED:
+            return State.CANCELLED
+        elif executor_state == ExecTaskState.FAILED:
+            return State.FAILED
+        elif executor_state == ExecTaskState.SUCCESS:
+            return State.FINISHED
+        else:
+            return State.UNKNOWN
+
+    def _write_script(self, ws_path, step):
+        """
+        Write a script to the workspace of a workflow step.
+
+        The job_map optional parameter is a map of workflow step names to job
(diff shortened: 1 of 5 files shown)


FrankD412:
@jwhite242 -- Some considerations based on why we were discussing earlier:
- I was hoping to eliminate the checks for local execution in the `ExecutionGraph._StepRecord`. This would bring all execution in line with scheduler-based adapters. You had mentioned some benefits to keeping the distinction around. What are your thoughts?
- We need to conceptualize how we would handle job packing in an Executor for pyaestro and then do something similar here. What have you found in terms of packing and the like?

FrankD412:
Current todo: Cancellation works at the Maestro level, but leaves the pool executing the last processes that were running before the cancellation was posted. This bug boils up to a pyaestro bug.

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+from maestrowf.abstracts.enums import JobStatusCode, State, SubmissionCode, \
+    CancelCode
+from maestrowf.interfaces.script import CancellationRecord, SubmissionRecord
+from pyaestro.utilities.executor import Executor, ExecTaskState, ExecCancel
+
+LOGGER = logging.getLogger(__name__)
+
+
+class LocalPoolAdapter(SchedulerScriptAdapter):
+    """Interface class for the flux scheduler (on Spectrum MPI)."""
```
jwhite242, line 14: Copy pasted doc string?

jwhite242: Actually, scrolling down more, is this whole thing really aimed at using with flux specifically?

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        is instantiated are as follows:
+        * num_processes: The number of local workers for processing.
+
+        :param **kwargs: A dictionary with default settings for the adapter.
+        """
+        super(LocalPoolAdapter, self).__init__(**kwargs)
+        num_workers = kwargs.get("num_workers", "1")
+        self.add_batch_parameter("max_workers", num_workers)
+        self._pool = Executor(num_workers)
+        self._extension = ".lp.sh"
```
jwhite242, line 32: so, if the plan is to just replace the base local adapter completely with these pool based critters, not sure we need the `.lp` qualifier here

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def extension(self):
```
jwhite242, line 35: What's the use case for this property? -> doc string

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        Generate the header present at the top of Flux execution scripts.
+
+        :param step: A StudyStep instance.
+        :returns: A string of the header based on internal batch parameters and
+                  the parameter step.
+        """
+        return f"#!{self._exec}"
+
+    def get_parallelize_command(self, procs, nodes=None, **kwargs):
```
jwhite242, line 48: Should we generalize this to execute or launch command rather than making parallel special?   More just about a hook for a custom launch prefix/suffix/wrapper/etc than parallel in particular right?  One scheduler thing i can see used is big memory jobs on hpc might request a whole node just for the memory, but only run serial (e.g. srun -N1 -n1..).

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        :param path: Local path to the script to be executed.
+        :param cwd: Path to the current working directory.
+        :param job_map: A dictionary mapping step names to their job
+                        identifiers.
+        :param env: A dict containing a modified environment for execution.
+        :returns: The return status of the submission command and job
+                  identifier.
+        """
+        try:
+            job_id = self._pool.submit(path, cwd)
```
jwhite242, line 74: there a future being tracked in here somewhere?

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+            LOGGER.debug("Specified parameter is not a list.")
+            if isinstance(joblist, int):
+                LOGGER.debug("Integer found.")
+                joblist = [joblist]
+            else:
+                LOGGER.debug("Unknown type. Returning an error.")
+                return JobStatusCode.ERROR, {}
+
+        status = {jid: State.UNKNOWN for jid in joblist}
+        try:
```
jwhite242, line 111: Would it be better to put this inside the loop so it can actually get through the whole list and catch all the exceptions?  Also, what exception did this get added to catch?

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        if not joblist:
+            return CancelCode.OK
+
+        c_status = set()
+        for job in joblist:
+            c_status.add(self._pool.cancel(job))
+
+        c_return = CancelCode.OK
+        ret_code = 0
+        if ExecCancel.FAILED in c_status:
```
jwhite242, line 140: Should this check all of them, i.e. counting the number of cancels that failed and ones that succeeded separately for more informative error messages?

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        :returns: A Study.State enum corresponding to parameter job_state.
+        """
+        if executor_state == ExecTaskState.PENDING:
+            return State.PENDING
+        elif executor_state == ExecTaskState.RUNNING:
+            return State.RUNNING
+        elif executor_state == ExecTaskState.INITIALIZED:
+            return State.INITIALIZED
+        elif executor_state == ExecTaskState.CANCELLED:
+            return State.CANCELLED
```
jwhite242, line 162: What about a CLOSING or CANCELLING status?  I've noticed at least on slurm that if you check it again too quick after cancelling (sometimes up to a minute) you'll get the CG status from the scheduler while it's still cleaning up/shutting it down.  Or can pending capture this?

samples/lulesh/lulesh_sample1_pool_macosx.yaml
```diff
@@ -0,0 +1,90 @@
...
+      git:
+        - name: LULESH
+          path: $(OUTPUT_PATH)
+          url: https://github.com/LLNL/LULESH.git
+          tag: 2.0.3
+
+batch:
+    shell: /bin/bash
+    type: local_pool
+    num_workers: 3
```
jwhite242, line 28: we'll definitely need to spend some effort on whatever nomenclature is settled on for this stuff to avoid confusion with workers/cores/tasks/jobs/...

maestrowf/interfaces/script/localpooladapter.py
FrankD412: Oh, that was totally a copy-paste mistake... I used the Flux adapter as a baseline for the API. I'll go back and fix that.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: Yeah -- that was just temporary to make sure I was actually generating what I thought since the local adapter is still there.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: I was thinking of refactoring this to be in base class API (assuming it isn't at the moment). And yeah, agreed; needs a docstring.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: This pool is entirely backed by `pyaestro` -- which does implement this with a future.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: In the current implementation, those don't exist. The pool technically will cancel a job (in this case a process) outright.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: This design choice is a decision carried forward from SLURM (when I initially made this); in that perspective, the status is reported **if at least one** record reports the state. The issue with things like SLURM is that it's more efficient to do a bulk cancel, so there's no way to get a per job report out there (not without polling the scheduler to see if it took).

maestrowf/interfaces/script/localpooladapter.py
FrankD412: I did this outside the loop since it's a basic operation -- also because the `Enum` is not a constructor, I couldn't use `defaultdict` like I wanted. I plan to explore making a meta-enum to handle the default case to make this more Pythonic.

FrankD412: In regards to the `except` -- it's there to log any possible errors that do come up. It's not necessarily meant to take corrective action.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: Yeah -- I want to redesign this in `pyaestro` and then use it here.

maestrowf/interfaces/script/localpooladapter.py
```diff
@@ -0,0 +1,204 @@
...
+        if not isinstance(joblist, list):
+            LOGGER.debug("Specified parameter is not a list.")
+            if isinstance(joblist, int):
+                LOGGER.debug("Integer found.")
+                joblist = [joblist]
+            else:
+                LOGGER.debug("Unknown type. Returning an error.")
+                return JobStatusCode.ERROR, {}
+
+        status = {jid: State.UNKNOWN for jid in joblist}
```
kcathey, line 110: Is `jid` here the same as `job_id` used in other places in this class? As I was reading I was confused for a moment and had to think about it, switching it to be `job_id` would preserve clarity even in a generator.

kcathey (changes requested):

jwhite242:
> @jwhite242 -- Some considerations based on why we were discussing earlier:
> 
>     * I was hoping to eliminate the checks for local execution in the `ExecutionGraph._StepRecord`. This would bring all execution in line with scheduler-based adapters. You had mentioned some benefits to keeping the distinction around. What are your thoughts?
> 
>     * We need to conceptualize how we would handle job packing in an Executor for pyaestro and then do something similar here. What have you found in terms of packing and the like?

So I was thinking this could be a useful distinction based on the two possible run modes: standalone batch jobs (the current scheduler adapters) and running maestro inside of an allocation manually for job packing (or spawned by another tool built on top of maestro).  Doesn't necessarily need to use local as the distinction, but do we need some hook to enable the latter job packing mode? 

As for the job packing, this is something where i think plugins would be really useful, or some way for users to write these things like with pgen.  The number of scheduling behaviors and the optimization algorithms for implementing them is a pretty large space that doesn't necessarily need to be hard wired into maestro.  A few simple/interesting ones would be good of course.

samples/lulesh/lulesh_sample1_pool_macosx.yaml
FrankD412: Should I rename this to num_processes instead? I feel like that would be a better description that doesn't alias to other things.

maestrowf/interfaces/script/localpooladapter.py
FrankD412: Will be done in my next commit.

FrankD412:
> > @jwhite242 -- Some considerations based on why we were discussing earlier:
> > ```
> > * I was hoping to eliminate the checks for local execution in the `ExecutionGraph._StepRecord`. This would bring all execution in line with scheduler-based adapters. You had mentioned some benefits to keeping the distinction around. What are your thoughts?
> > 
> > * We need to conceptualize how we would handle job packing in an Executor for pyaestro and then do something similar here. What have you found in terms of packing and the like?
> > ```
> 
> So I was thinking this could be a useful distinction based on the two possible run modes: standalone batch jobs (the current scheduler adapters) and running maestro inside of an allocation manually for job packing (or spawned by another tool built on top of maestro). Doesn't necessarily need to use local as the distinction, but do we need some hook to enable the latter job packing mode?
> 
> As for the job packing, this is something where i think plugins would be really useful, or some way for users to write these things like with pgen. The number of scheduling behaviors and the optimization algorithms for implementing them is a pretty large space that doesn't necessarily need to be hard wired into maestro. A few simple/interesting ones would be good of course.

I was thinking about this and the case of running Maestro in an allocation would be where you have an "allocation adapter" class that comes in handy. That class would take in the global set of resources requested and schedule the conductor call. That's actually where I wanted to split out the MPI related functionality from the `SchedulerScriptAdapter` since you could either monkey patch the method in the `LocalPoolAdapter` to generate the right MPI or just have it in the batch settings to use a particular MPI (usual factory pattern there to get the class). We could mock that up in a discussion thread.

For this PR, I was thinking if I can get the local pool as the main local adapter (it would still return that it's local, but in the `ExecutionGraph` it'd be treated just like everything else since it's a process pool. The records would still return local otherwise. 

I'm starting to wonder if this is a good time to introduce an `orchestrate` sub-command to handle the allocation clear so it's more transparent to the user? -- we should make a discussion on this haha

Hopefully this is making sense.

PR #352, opened 2021-03-03
Spectrum adapter remove/lingering in-class import.

FrankD412:
When using the `FluxScriptAdapter` with MuMMI, the adapter failed to pickle when used the multiprocessing libraries. The adapter contained a stale `__import__` statement which caused `cPickle` to throw an exception when attempting to pickle the module. While cleaning up the stale import, the opportunity was timely to remove the old and now deprecated `SpectrumFluxScriptAdapter` which is no longer used anywhere.

Short term: This fixes a bug where a stray import prevented the
FluxScriptAdapter from being pickled for interprocess communication.

Long term: Cleans up the code from the stale Spectrum adapter and
makes it so that we have less dead code. It prevents us having to
waste time looking through code that is no longer used.


maestrowf/interfaces/script/_flux/flux0_18_0.py
```diff
@@ -11,7 +11,7 @@
 LOGGER = logging.getLogger(__name__)
 
 try:
-    import flux
+    from flux import constants, Flux, job
```
jwhite242, line 14: Are job/constants maybe too generic here?  want a flux or f prefix or something on these? Those seem ripe for usage elsewhere in maestro..

FrankD412: You're right -- I ended up having to go through the adapter and update local variables that aliased with these. These are local to this module though, so I don't have much concern of them bleeding out elsewhere. I can tidy these up though.

jwhite242: yeah, more wondering about bleeding into this module from outside, not the other way

jwhite242 (approved):

PR #353, opened 2021-03-08
Draft: Update pip and tox for 3.6

FrankD412:
This PR is trying to fix the new 3.6 `cryptography` error that's been popping up.


FrankD412:
Solution found here: https://github.com/python-poetry/poetry/issues/3661#issuecomment-775500018

PR #357, opened 2021-05-06
Flux update

SteVwonder:
Fixes and updates based on the GMD loop work.  This could probably use some testing since the 0.26.0 adaptor was originally based off of a different base branch.  Let me know how you'd like to go about testing @FrankD412.  I'm happy to help setup some flux-based CI testing if you are interested.

### DIFF
--- a/docker/flux/0.26.0/Dockerfile
+++ b/docker/flux/0.26.0/Dockerfile
@@ -1,0 +1,19 @@
+FROM fluxrm/flux-core:centos7-v0.26.0
+ENV PATH="/home/fluxuser/.local/bin:$PATH"
+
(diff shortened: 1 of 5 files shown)


FrankD412:
@SteVwonder -- firstly, thanks for posting this PR! I think we'll want to keep 0.18.0 the way it was since I've been trying to leave interfaces available for older versions to allow users to use them. So I do think we'll want to bring back the 0.26.0 interface module.

For testing, yes please -- I do need to move to GitHub actions by end of May, so this might be a good way to start. I'm not sure how we can test Flux as I'm not the most familiar with the full range of testing capabilities; would be happy to sit down and talk about how to do it if you'd like.

PR #358, opened 2021-05-07
Status tweaks

jwhite242:
Adds some quality of life tweaks to the status outputs:

- makes step order deterministic, using breadth first search ordering
- uses rich library to improve table formatting and make things more readable
- adds layout switch to better accommodate long step and workspace names and improve view in narrow terminals
  - --layout cli arg, with 'flat' (default) and 'narrow' options currently

Sample outputs of flat and narrow layout switches:

Flat (default):
image

Narrow:
image

Note, these colors will vary with terminal them: current snapshots from the encom theme in iterm2


jwhite242:
So, one other potential feature to add to this: adding tabulated parameters and their values for each one.  This would be easiest to do in the narrow layout with the nested table setup, but could probably build an alternate flat view too (leave this flat view alone and add a new layout).  Any interest in adding that to this PR?

FrankD412:
@jwhite242 -- I'd be interested in seeing that in this PR if you're willing to give it a go.

FrankD412:
I'm seeing that `rich` is unsatisfiable. I'm double checking the python version we're requesting in develop, but just wanted to post the FYI.

```
 The current project's Python requirement (>3.5) is not compatible with some of the required packages Python requirement:
    - rich requires Python >=3.6,<4.0, so it will not be satisfied for Python >3.5,<3.6 || >=4.0
```

jwhite242:
That's interesting.  Given that our CI tests are 3.6+ already, should 3.5 just be retired?

FrankD412:
Ok -- your branch does match with develop. I believe that 3.5 is now deprecated, with 3.6 being deprecated at the end of the year. I'm wondering if I shouldn't bump the version up in develop.

FrankD412:
> That's interesting. Given that our CI tests are 3.6+ already, should 3.5 just be retired?

Yep -- exactly what I was thinking.

kcathey:
> > That's interesting. Given that our CI tests are 3.6+ already, should 3.5 just be retired?
> 
> Yep -- exactly what I was thinking.

I'd agree as well.

maestrowf/maestro.py
```diff
@@ -72,20 +98,146 @@ def status_study(args):
...
+                            row_style = 'row_style'
+
+                        step_sched.add_row(sched_row,
+                                           status[sched_row][row],
+                                           style=row_style)
+
+                    step_details.add_row(step_info, step_sched)
+
+                    step_table.add_row('', step_details)
+                    stat_table.add_row(step_table, end_section=True)
```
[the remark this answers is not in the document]
FrankD412: It might be worth having a factory for the different table functions.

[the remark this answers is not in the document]
FrankD412: Actually, I think I did this somewhere in one of the command line args.. you can have the factory return the valid choices for the `choices=` parameter in `add_argument`. Then when any new options are added, it automatically gets updated.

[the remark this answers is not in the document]
jwhite242: oh, that sounds much nicer.  layout factory located elsewhere it is

FrankD412:
Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.

FrankD412:
@jwhite242 -- When I tested this on a previous LULESH study I got the following; haven't root caused it yet.

```
(maestrowf-hgfr9oqT-py3.7)  frank@Friday  ~/code/Python/maestro/maestrowf   status_tweaks  maestro status --layout narrow ./sample_output/lulesh/lulesh_sample1_20210413-090525
STATUS keys: ['Step Name', 'Job ID', 'Workspace', 'State', 'Run Time', 'Elapsed Time', 'Start Time', 'Submit Time', 'End Time', 'Number Restarts']
  File "/home/frank/.cache/pypoetry/virtualenvs/maestrowf-hgfr9oqT-py3.7/bin/maestro", line 5, in <module>
    main()
  File "/mnt/c/Users/frank/Documents/Code/Python/maestro/maestrowf/maestrowf/maestro.py", line 674, in main
    rc = args.func(args)
  File "/mnt/c/Users/frank/Documents/Code/Python/maestro/maestrowf/maestrowf/maestro.py", line 230, in status_study
    param_list = status['Params'][row].split(';')
KeyError: 'Params'
```

jwhite242:
> @jwhite242 -- When I tested this on a previous LULESH study I got the following; haven't root caused it yet.
> 
> ```
> (maestrowf-hgfr9oqT-py3.7)  frank@Friday  ~/code/Python/maestro/maestrowf   status_tweaks  maestro status --layout narrow ./sample_output/lulesh/lulesh_sample1_20210413-090525
> STATUS keys: ['Step Name', 'Job ID', 'Workspace', 'State', 'Run Time', 'Elapsed Time', 'Start Time', 'Submit Time', 'End Time', 'Number Restarts']
> Traceback (most recent call last):
>   File "/home/frank/.cache/pypoetry/virtualenvs/maestrowf-hgfr9oqT-py3.7/bin/maestro", line 5, in <module>
>     main()
>   File "/mnt/c/Users/frank/Documents/Code/Python/maestro/maestrowf/maestrowf/maestro.py", line 674, in main
>     rc = args.func(args)
>   File "/mnt/c/Users/frank/Documents/Code/Python/maestro/maestrowf/maestrowf/maestro.py", line 230, in status_study
>     param_list = status['Params'][row].split(';')
> KeyError: 'Params'
> ```

It perturbs what gets written to the status file to make this work, so you'll have to run it on new ones, but that points out a bug: this def isn't backwards compatible in it's current form.  Should be easy to fix that

maestrowf/maestro.py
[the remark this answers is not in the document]
FrankD412: What I normally do is put the factory classes in the `__init__` files for the module "above" the items being controlled, but that's really useful for things like adapters and other interface items. In this case, since it's part of the front end of Maestro, I could see it being put in `maestro.py` or in the top level `__init__` file for general use. To me this is UI specific to Maestro.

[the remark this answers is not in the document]
jwhite242: alright, sounds good; just wasn't sure if you wanted to clutter up the main file.  i'll stick it in there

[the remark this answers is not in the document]
jwhite242: Ok, given the filtering tools are gonna need to live in utils, think the concrete renderers and their registration is going to need to move into maestro.py so the filters can actually be imported.

FrankD412 (dismissed):

jwhite242:
> Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.

So, should this be stored separately in the step record then rather than trying to tack it onto the records workspaces?  (assuming tacking it on wouldn't break things)

FrankD412:
> > Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.
> 
> So, should this be stored separately in the step record then rather than trying to tack it onto the records workspaces? (assuming tacking it on wouldn't break things)

It's actually already stored [here](https://github.com/LLNL/maestrowf/blob/develop/maestrowf/datastructures/core/executiongraph.py#L50). It might be worth making a property to be able to get it easily.

FrankD412:
> > > Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.
> > 
> > 
> > So, should this be stored separately in the step record then rather than trying to tack it onto the records workspaces? (assuming tacking it on wouldn't break things)
> 
> It's actually already stored [here](https://github.com/LLNL/maestrowf/blob/develop/maestrowf/datastructures/core/executiongraph.py#L50). It might be worth making a property to be able to get it easily.
> 
> Yeah, but, it looks like it's already writing that value to the status file?

You're right actually, my bad. It's [here](https://github.com/LLNL/maestrowf/blob/bce8baccb24af350b821847a4dd87edd34365d55/maestrowf/datastructures/core/executiongraph.py#L617). It should probably be something different than the `os.path.split` to grab the last two entries.

jwhite242:
> > > > Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.
> > > 
> > > 
> > > So, should this be stored separately in the step record then rather than trying to tack it onto the records workspaces? (assuming tacking it on wouldn't break things)
> > 
> > 
> > It's actually already stored [here](https://github.com/LLNL/maestrowf/blob/develop/maestrowf/datastructures/core/executiongraph.py#L50). It might be worth making a property to be able to get it easily.
> > Yeah, but, it looks like it's already writing that value to the status file?
> 
> You're right actually, my bad. It's [here](https://github.com/LLNL/maestrowf/blob/bce8baccb24af350b821847a4dd87edd34365d55/maestrowf/datastructures/core/executiongraph.py#L617). It should probably be something different than the `os.path.split` to grab the last two entries.

So, would it be best to grab this [here](https://github.com/LLNL/maestrowf/blob/bce8baccb24af350b821847a4dd87edd34365d55/maestrowf/datastructures/core/study.py#L664) during study construction and attach it (and perhaps unexpanded step name?) to step records separately?

FrankD412:
> > > > > Something else I noticed, in the `narrow` style -- the printed workspace is just the subdirectory. You would likely want to have the step directory in front of that for an actual path.
> > > > 
> > > > 
> > > > So, should this be stored separately in the step record then rather than trying to tack it onto the records workspaces? (assuming tacking it on wouldn't break things)
> > > 
> > > 
> > > It's actually already stored [here](https://github.com/LLNL/maestrowf/blob/develop/maestrowf/datastructures/core/executiongraph.py#L50). It might be worth making a property to be able to get it easily.
> > > Yeah, but, it looks like it's already writing that value to the status file?
> > 
> > 
> > You're right actually, my bad. It's [here](https://github.com/LLNL/maestrowf/blob/bce8baccb24af350b821847a4dd87edd34365d55/maestrowf/datastructures/core/executiongraph.py#L617). It should probably be something different than the `os.path.split` to grab the last two entries.
> 
> So, would it be best to grab this [here](https://github.com/LLNL/maestrowf/blob/bce8baccb24af350b821847a4dd87edd34365d55/maestrowf/datastructures/core/study.py#L664) during study construction and attach it (and perhaps unexpanded step name?) to step records separately?

Probably -- I think that @ckrenn was introducing something similar by adding it directly to the `StudyStep` class (he calls it a `step_label`). It may be worth adding it as a `basename` but making a property for it so that it's actually part of the API. Each `ExecutionGraph._StepRecord` contains the `StudyStep` as a reference. That's probably very memory inefficient, and given the chance to go back I'd make it so that it could reference an unexpected copy -- but for now that is what it is.

kcathey (approved):

FrankD412:
@jwhite242 -- Just a reminder, this is still printing some debug:

```
Setting 'Step Name' column to style 'Step Name'
Setting 'Job ID' column to style 'Job ID'
Setting 'Workspace' column to style 'Workspace'
Setting 'State' column to style 'State'
```

jwhite242:
> @jwhite242 -- Just a reminder, this is still printing some debug:
> 
> ```
> Setting 'Step Name' column to style 'Step Name'
> Setting 'Job ID' column to style 'Job ID'
> Setting 'Workspace' column to style 'Workspace'
> Setting 'State' column to style 'State'
> ```

Good catch!  They're all gone now

maestrowf/__init__.py
```diff
@@ -53,3 +60,372 @@ def emit(self, record):
...
+        renderer = self._layouts.get(layout)
+
+        # Note, need to wrap renderer in try/catch too, or return default val?
+        if not renderer:
+            raise ValueError(layout)
+
+        return renderer()
+
+
+class BaseStatusRenderer:
```
FrankD412, line 84: It looks like this should be an abstract class and the `layout` and `render` methods should be `abstractmethods` -- see the `SchedulerScriptAdapter` for an example.

maestrowf/__init__.py
```diff
@@ -53,3 +60,372 @@ def emit(self, record):
 
 __version_info__ = ("1", "1", "9dev0")
 __version__ = '.'.join(__version_info__)
+
+
+# Register status layout types
+class StatusRendererFactory:
```
FrankD412, line 66: There's two other factories that you could model one after -- almost wondering if it shouldn't just be made into a metaclass or something -- but those two base their keys on grabbing the classes of a specific base class. Another suggestion would be to add a method for getting the available renderers and then having it return a list of keys. That's useful in the `parser.add_argument(..., 'choices=StatusRendererFactory.keys( )')` optional so that the argument gets updated automatically.

maestrowf/__init__.py
```diff
@@ -53,3 +60,372 @@ def emit(self, record):
...
+            "Step Name": "bold",
+            "Workspace": "blue",
+            "row_style": "",
+            "row_style_dim": "dim",
+            "background": "grey7"
+        }
+
+    def layout(self, status_data=None, study_title=None, filter_dict=None):
+
+        if status_data:
```
FrankD412, line 255: In this case, just drop the default arguments -- if data is required, it might be worth checking that it matches a type instead of defaulting it to none.

maestrowf/maestro.py
```diff
@@ -436,6 +445,10 @@ def setup_argparser():
     status.add_argument(
         "directory", type=str, nargs="+",
         help="Directory containing a launched study.")
+    status.add_argument(
+        "--layout", type=str, choices=['flat', 'narrow', 'legacy'],
```
FrankD412, line 449: I believe this will error out if not in the choices and is automatically handled by the argument parser class.

FrankD412 (changes requested):

maestrowf/__init__.py
jwhite242: oops, my bad! definitely what this should be

maestrowf/__init__.py
jwhite242: Ok, got this one and the other converted to use abstract base objects and got the auto-registration and cli population working.  This variant still keeps it in the __init__, but let me know if you'd prefer pushing the abstract base and concrete classes into submodules as with the script adapter setup.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -65,6 +65,19 @@ def __init__(self, workspace, step, **kwargs):
         self._end_time = None
         self.status = kwargs.get("status", State.INITIALIZED)
 
+        # Parameter info
+        self._params = None
+
+    def add_params(self, params):
```
FrankD412, line 71: Documenting `params` appears to be missing.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -392,6 +408,9 @@ def add_step(self, name, step, workspace, restart_limit):
                     "restart_limit": restart_limit,
                 }
         record = _StepRecord(**data)
+        if params:
+            record.add_params(params)
```
FrankD412, line 412: Should this be a hidden method?

maestrowf/maestro.py
```diff
@@ -67,25 +67,35 @@ def status_study(args):
...
-            print(abs_path)
-            print(header_format)
-
             status = Conductor.get_status(abs_path)
+            status_layout = args.layout
+
             if status:
-                print(tabulate.tabulate(status, headers="keys"))
+                try:
+                    # Wasteful to not reuse this renderer for all paths?
```
FrankD412, line 79: In terms of wasteful, probably; but, for the scales that we're at I don't think this is a limiting factor.

maestrowf/datastructures/core/executiongraph.py
jwhite242: Uh, don't think so? That would suggest this sort of external access shouldn't be allowed (i know python doesn't really block that..).

maestrowf/datastructures/core/executiongraph.py
jwhite242: Got both missing instances of this; also patched up Combo's param values doctstring while tracing what the type of this critter was.

maestrowf/datastructures/core/executiongraph.py
FrankD412: Got it, I see what you're saying -- I think I got confused while skimming the changes, but it's a method on the record.

maestrowf/__init__.py
```diff
@@ -53,3 +65,472 @@ def emit(self, record):
...
+                                    show_header=False,
+                                    show_lines=True,
+                                    box=box.HORIZONTALS)
+
+                # Note col names don't actually matter, just setting styles
+                step_params.add_column("name", style="")
+                step_params.add_column("val", style="blue", justify="right")
+                step_params.add_column("name2", style="")
+                step_params.add_column("val2", style="blue", justify="right")
+                print("Param list: {}".format(param_list))
```
FrankD412, line 419: Just a reminder -- the debug printing is back.

jwhite242: Whoops!  my bad

FrankD412:
@jwhite242 -- Were you planning to change the documentation to call out legacy, or would you consider this PR done?

jwhite242:
> @jwhite242 -- Were you planning to change the documentation to call out legacy, or would you consider this PR done?

Yeah, i'll re-add the legacy layout in the documentation shortly, then i'm done.  and feel free to squash merge all of it if you don't want any of the questionable debugging junk left in the history (or request i squash those out, that's fine too)

jwhite242:
Also, just to keep it in the discussion, here's the test results for what happens with many parameters, enough that hashing is required to avoid max file name length errors:
image

And here's the same thing with the nominal flat layout:
image

Note, we could add an option to truncate the step names in such cases instead of wrapping them, with the hashed workspace still being visible to let users drill down into each one.  Only side effect is the very tall rows.

FrankD412 (approved):

PR #359, opened 2021-05-18
Flux update

SteVwonder:
This PR adds several changes to the 0.26.0 adapter based on experiences with the GMD loop.  In particular:

- set the default value for `force_broker` to `True` to maximize compatibility with other schedulers/backends
- handle errors in the type for `ngpu` in the flux adapter
- flux the number of slots used in the nested Flux case
- add version checking and log an error when the versions (broker vs adapter) are incompatible

I confirmed that the extra logging does not trigger when running 0.26.0 against 0.26.0 and the error is displayed when running against an older Flux (I had 0.19.0 on hand on Pascal):

```
[2021-05-17 21:08:51: DEBUG] [flux0_26_0: 28] New Flux handle created.
[2021-05-17 21:08:51: DEBUG] [flux0_26_0: 32] Connected to Flux broker running version 0.19.0 using maestro adaptor version 0.26.0.
[2021-05-17 21:08:51: ERROR] [flux0_26_0: 42] Maestro adaptor version (0.26) is too new for the Flux broker version (0.19). Functionality not present in this Flux v
ersion may be required by the adaptor and cause errors. Please switch to an older adaptor.
```

### DIFF
--- a/maestrowf/interfaces/script/_flux/flux0_26_0.py
+++ b/maestrowf/interfaces/script/_flux/flux0_26_0.py
@@ -64,6 +64,9 @@ class FluxInterface_0260(FluxInterface):
         # single node, don't use a broker -- but introduce a flag that can
         # force a single node to run in a broker.
 
+        if isinstance(ngpus, str) and ngpus.isdigit():
+            ngpus = int(ngpus)
+
         if force_broker or nodes > 1:
             LOGGER.debug(
                 "Launch under Flux sub-broker. [force_broker=%s, nodes=%d]",
--- a/maestrowf/interfaces/script/_flux/flux0_26_0.py
+++ b/maestrowf/interfaces/script/_flux/flux0_26_0.py
@@ -22,39 +22,6 @@ class FluxInterface_0260(FluxInterface):
     flux_handle = None
 
     @classmethod
-    def connect_to_flux(cls):
-        if not cls.flux_handle:
-            cls.flux_handle = flux.Flux()
-            LOGGER.debug("New Flux handle created.")
-            broker_version = cls.flux_handle.attr_get("version")
-            adaptor_version = cls.key
-            LOGGER.debug(
-                "Connected to Flux broker running version %s using Maestro "
-                "adapter version %s.", broker_version, adaptor_version)
-            try:
-                from distutils.version import StrictVersion
-                adaptor_version = StrictVersion(adaptor_version)
-                broker_version = StrictVersion(broker_version)
-                if adaptor_version > broker_version:
-                    LOGGER.error(
-                        "Maestro adapter version (%s) is too new for the Flux "
-                        "broker version (%s). Functionality not present in "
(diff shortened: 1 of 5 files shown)


maestrowf/interfaces/script/_flux/flux0_26_0.py
```diff
@@ -21,14 +21,45 @@ class FluxInterface_0260(FluxInterface):
 
     flux_handle = None
 
+    @classmethod
+    def connect_to_flux(cls):
+        if not cls.flux_handle:
+            cls.flux_handle = flux.Flux()
+            LOGGER.debug("New Flux handle created.")
+            broker_version = cls.flux_handle.attr_get("version")
```
FrankD412, line 29: @SteVwonder -- Curiosity question, do all Flux versions support this or is this new? If so, maybe we should promote this to a higher order method and construct the handle in the front end class?

SteVwonder: It was added in this PR (https://github.com/flux-framework/flux-core/pull/1817) three years ago.  Thus it is in v0.10.0+.  So I think it is safe to promote this as you suggest.

maestrowf/interfaces/script/_flux/flux0_26_0.py
```diff
@@ -37,6 +35,9 @@ def submit(
         # single node, don't use a broker -- but introduce a flag that can
         # force a single node to run in a broker.
 
+        if isinstance(ngpus, str) and ngpus.isdigit():
```
SteVwonder, line 38: Is this type/isdigit checking still required after the changes to the upper-level fluxscriptadaptor?

SteVwonder (commented):
Thanks @FrankD412 for getting these changes in.  Just two comments:

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -183,9 +183,14 @@ def submit(self, step, path, cwd, job_map=None, env=None):
                 "'cores per task' set to a non-value. Populating with a "
                 "sensible default. (cores per task = %d", cores_per_task)
 
-        # Calculate ngpus
-        ngpus = step.run.get("gpus", 0)
-        ngpus = 0 if not ngpus else ngpus
+        try:
+            # Calculate ngpus
+            ngpus = step.run.get("gpus", "0")
+            ngpus = int(ngpus) if ngpus else 0
```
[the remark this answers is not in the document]
FrankD412: In this case, this is due to the fact that a `StudyStep` defaults this value to `""` -- a revisit to the `StudyStep` is required to correct this.

maestrowf/interfaces/script/_flux/flux0_26_0.py
FrankD412: Good catch -- missed this one.

PR #365, opened 2021-06-17
Slurm QOS specification

### LINKED ISSUE #336: Add qos option to slurm adaptor
Allow specifying the `qos` parameter for slurm via the batch block.

```
batch:
    type: slurm
    ...
    qos: standby

```

Leading to the additional header line:
```
#SBATCH --qos=standby
```

youngjeffrey:
Fixes #336


FrankD412:
@youngjeffrey -- this looks fine, except you've checked in your renamed toml file. Please remove that from the commit. After that, I think this is good to go.

PR #394, opened 2022-05-15
Move to purely poetry install

FrankD412:
It looks like with the introduction of recent `poetry-core` updates, the core API now supports editable installs (see [here](https://github.com/python-poetry/poetry/issues/34#issuecomment-1054626460)). This will make it so that we no longer need to have a `setup.py` alongside the `pyproject.toml` in the repository.

This PR removes `setup.py` and adds the necessary versioning to the `pyproject.toml` file to make editable possible. It also adds basic install tests to make sure that when we make tweaks to installation that it does not break.


jwhite242 (commented):
Nice to see editable installs in poetry!

I am a little concerned about dropping setup.py given the number of users that use editable installs but don't use poetry (still lots of pip + venv and conda users that install from repos instead of wheels).  Is there a way to keep a dummy setup.py that setuptools can use as a fallback?  Something like what's at the top here: [setuptools pyproject config](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)
Additionally, the dynamic metadata for it at the bottom might be useful for the version?

One more question: what, if any implications are there from using poetry to manage envs with packages that have c/c++ dependencies that may need to also be installed in editable mode (perhaps even needing some baked into maestro itself in the future)?

jwhite242:
Oh, one more comment: version tick?

jwhite242:
Yet another thought: maybe time for a contributing section in the docs for at least mentioning the use of poetry (perhaps not a full tutorial, but maybe a quick start or something?)

FrankD412:
> Nice to see editable installs in poetry!
> 
> I am a little concerned about dropping setup.py given the number of users that use editable installs but don't use poetry (still lots of pip + venv and conda users that install from repos instead of wheels). Is there a way to keep a dummy setup.py that setuptools can use as a fallback? Something like what's at the top here: [setuptools pyproject config](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) Additionally, the dynamic metadata for it at the bottom might be useful for the version?

So, for the initial concern about dropping `setup.py` -- with the new implementations of the backend that's specified, this means that `pip install -e .` now works. This PR won't change the workflow for editable installs and doesn't require poetry to work. For the concern about classifiers, I can actually port those -- it looks like in the page you linked to, there's actually classifiers in their `pyproject.toml`.
 
> One more question: what, if any implications are there from using poetry to manage envs with packages that have c/c++ dependencies that may need to also be installed in editable mode (perhaps even needing some baked into maestro itself in the future)?

That's a good question -- I'll need to look into that one. It's not something I anticipate needing short term, so it might not be an issue by the time we need it, if we do. I have to look into it for another project anyway, probably -- so I'll be prepared for that hopefully.

> Yet another thought: maybe time for a contributing section in the docs for at least mentioning the use of poetry (perhaps not a full tutorial, but maybe a quick start or something?)

For this PR, it still allows the `pip install -e` flow -- I can definitely add a section to the docs about developing in Poetry since that's now an option.

jwhite242 (approved):
looks good to me pending the version tick.  thanks for cleaning this up!  even makes my conda workflows much less hacky!

PR #396, opened 2022-05-16
Ci/test and release

jwhite242:
Turns on running of unit tests against multiple python versions upon pushes and pull requests


jwhite242:
So, general strategy here is to be a little greedy with the unit tests and not require linting to succeed first -> ideally catch errors in both in one go.  Happy to change that though if nobody else likes that plan

FrankD412:
I'd prefer it fail linting -- I think it's a good incentive to have good style. We'll test your changes if you help us enforce style. Win-win in my book. 😝

FrankD412 (approved):

PR #404, opened 2022-06-28
Print usage when no args provided

### LINKED ISSUE #188: AttributeError: 'Namespace' object has no attribute 'func'
When running maestro you can get the following error.
I am running Python 3.7.2 and my pip version is 19.1.1

kinow:
Hi,

Running `maestro` for the first time ever today from the `develop` branch. I first tried `maestro` without anything else, but it resulted in a traceback.

```bash
(venv) kinow@ranma:~/Development/python/workspace/maestrowf$ maestro 
  File "/home/kinow/Development/python/workspace/maestrowf/venv/bin/maestro", line 8, in <module>
    sys.exit(main())
  File "/home/kinow/Development/python/workspace/maestrowf/maestrowf/maestro.py", line 495, in main
    if not hastattr(args, 'func'):
NameError: name 'hastattr' is not defined
```
I think it would be best to print the program usage instead? So if users are not aware, or not sure, which subcommand/subparser to trigger, they can read the usage and decide then.

Thanks for maestro! :wave: 
-Bruno

EDIT: closes #188 (hadn't seen the issue, sorry)

### DIFF
--- a/maestrowf/maestro.py
+++ b/maestrowf/maestro.py
@@ -369,6 +369,9 @@ def setup_argparser():
         description="The Maestro Workflow Conductor for specifying, launching"
         ", and managing general workflows.",
         formatter_class=RawTextHelpFormatter)
+    # This call applies a default function to the main parser as described
+    # here: https://stackoverflow.com/a/61680800
+    parser.set_defaults(func=lambda args: parser.print_help())
     subparsers = parser.add_subparsers(dest='subparser')
 
     # subparser for a cancel subcommand


FrankD412:
Hi @kinow -- Thanks for the contribution! Must say you have impeccable timing; I had someone encounter this recently, so it was on my mind.

I've approved the MR to run through our CI and will review the code in a bit. :)

Thanks again for your contribution!

FrankD412 (changes requested):
@kinow -- it looks like you can set the default function. I generally avoid `hasattr` since that's been considered bad practice from what I've seen. Talking with @jwhite242 and with some searching, it seems that this might be the better bet:

```
    parser = ArgumentParser(
        prog="maestro",
        description="The Maestro Workflow Conductor for specifying, launching"
        ", and managing general workflows.",
        formatter_class=RawTextHelpFormatter)
    # This call applies a default function to the main parser as described
    # here: https://stackoverflow.com/a/61680800
    parser.set_defaults(func=lambda args: parser.print_help())
```

kinow:
Hi @FrankD412 

I normally use `hasattr` or wrap it in a try-except block for a `AttributeError` (latter is normally faster IIRC?). But I have no preference, it was just the first idea that came to my mind.

+1 to using a default value if that's available. Give me 10 mins and I'll update the branch and test locally…

maestrowf/maestro.py
```diff
@@ -369,6 +369,9 @@ def setup_argparser():
         description="The Maestro Workflow Conductor for specifying, launching"
         ", and managing general workflows.",
         formatter_class=RawTextHelpFormatter)
+    # This call applies a default function to the main parser as described
+    # here: https://stackoverflow.com/a/61680800
+    parser.set_defaults(func=lambda args: parser.print_help())
```
kinow, line 374: Much cleaner! And worked like a charm. Thanks @FrankD412 :bow:

FrankD412: Of course, @kinow! Anytime! Thank you so much for submitting a fix. :-)

FrankD412 (approved):

PR #405, opened 2022-06-30
Bugfix/slurm job checking

jwhite242:
Fixes issue with job checking on slurm systems that can result in steps never being marked finished.  The squeue command appears to flush finished/cancelled/killed jobs from the queue before maestro can check on the jobs and update their status.  Have yet to reproduce this in maestro studies, but can reproduces with manual job submissions and using ctrl-c to kill them and watching the squeue iterate output drop the job almost immediately.  This fix replaces squeue with sacct which preserves the job info for much longer.

### DIFF
--- a/maestrowf/interfaces/script/slurmscriptadapter.py
+++ b/maestrowf/interfaces/script/slurmscriptadapter.py
@@ -245,25 +245,117 @@ class SlurmScriptAdapter(SchedulerScriptAdapter):
                 "Submission returned an error (see next line).\n%s", err)
             return SubmissionRecord(SubmissionCode.ERROR, retcode)
 
-    def check_jobs(self, joblist):
+    def _check_jobs_squeue(self, joblist, status):
+        """
+        For the given job list, query execution status.
+        This method uses squeue command to query the scheduler and does a
+        regex search for job information.
+        :param joblist: A list of job identifiers to be queried.
+        :param status: Dictionary of jobid:job status to fill out
+        :returns: The return code of the status query, status dictionary
+        """
+        # squeue options:
+        # -u = username to search queues for.
+        # -t = list of job states to search for. 'all' for all states.
+
+        # The squeue command output is split with the following indices
+        # used for specific information:
+        # 0 - Job Identifier
+        # 1 - Queue
+        # 2 - Job name
+        # 3 - User
+        # 4 - State [Passed to _state]
+        # 5 - Current Execution Time
+        # 6 - Assigned Node Count
+        # 7 - Hostname and assigned node identifier list
+
+        cmd = "squeue -u $USER -t all"
+
+        # Indices of needed columns in squeue output
+        data_row_offset = 1     # Just header, no header/row separator
+        state_index = 4
+        jobid_index = 0
+
+        # status = {}
+        # for jobid in joblist:
+        #     LOGGER.debug("Looking for jobid %s", jobid)
+        #     status[jobid] = None
+
+        p = start_process(cmd)
+        output, err = p.communicate()
+        retcode = p.wait()
+
+        if retcode == 0:    # Successfully checked scheduler, parse output
+            for job in output.split("\n")[data_row_offset:]:
+                LOGGER.debug("Job Entry: %s", job)
+
+                job_split = re.split(r"\s+", job)
+
+                # Check for blank entry in first column
+                if job_split[0] == "":
+                    LOGGER.debug("Removing blank entry from head of status.")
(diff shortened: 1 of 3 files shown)


jwhite242:
Screen shot for documenting the default outputs from running sacct on the hello_bye_parameterized spec:

```
sacct --jobs=6564055,6564056,6564057,6564058,6564059,6564060,6564061,6564062
JobID           JobName  Partition    Account  AllocCPUS      State ExitCode 
------------ ---------- ---------- ---------- ---------- ---------- -------- 
6564055      hello_wor+     pdebug    wbronze          1  COMPLETED      0:0 
6564055.bat+      batch               wbronze          1  COMPLETED      0:0 
6564055.0          echo               wbronze          1  COMPLETED      0:0 
6564056      hello_wor+     pdebug    wbronze          1  COMPLETED      0:0 
6564056.bat+      batch               wbronze          1  COMPLETED      0:0 
6564056.0          echo               wbronze          1  COMPLETED      0:0 
6564057      hello_wor+     pdebug    wbronze          1  COMPLETED      0:0 
6564057.bat+      batch               wbronze          1  COMPLETED      0:0 
6564057.0          echo               wbronze          1  COMPLETED      0:0 
6564058      hello_wor+     pdebug    wbronze          1  COMPLETED      0:0 
6564058.bat+      batch               wbronze          1  COMPLETED      0:0 
6564058.0          echo               wbronze          1  COMPLETED      0:0 
6564059      bye_world+     pdebug    wbronze          1  COMPLETED      0:0 
6564059.bat+      batch               wbronze          1  COMPLETED      0:0 
6564059.0          echo               wbronze          1  COMPLETED      0:0 
6564060      bye_world+     pdebug    wbronze          1  COMPLETED      0:0 
6564060.bat+      batch               wbronze          1  COMPLETED      0:0 
6564060.0          echo               wbronze          1  COMPLETED      0:0 
6564061      bye_world+     pdebug    wbronze          1  COMPLETED      0:0 
6564061.bat+      batch               wbronze          1  COMPLETED      0:0 
6564061.0          echo               wbronze          1  COMPLETED      0:0 
6564062      bye_world+     pdebug    wbronze          1  COMPLETED      0:0 
6564062.bat+      batch               wbronze          1  COMPLETED      0:0 
6564062.0          echo               wbronze          1  COMPLETED      0:0
```

The machinery used to split lines currently ignores the job steps and simply uses the cumulative job state.

maestrowf/interfaces/script/slurmscriptadapter.py
```diff
@@ -344,21 +353,24 @@ def _state(self, slurm_state):
         :returns: A Study.State enum corresponding to parameter job_state.
         """
         LOGGER.debug("Received SLURM State -- %s", slurm_state)
-        if slurm_state == "R":
+        if slurm_state == "R" or slurm_state == "RUNNING":
```
FrankD412, line 356: Can these actually be abbreviated in `sacct`?

FrankD412:
@jwhite242 -- For not being able to test this directly, these changes look like they're just fine. Just one question, but otherwise

FrankD412:
Random thought, might it be worth checking if the call to `sacct` fails and then backing off to `squeue`?

jwhite242:
> 

Yeah, I could go back to that if you prefer; initial implementation of this used a fallback to sacct when squeue failed and kept the same end check for both of them losing track of jobs.   One other related hook that might be interesting here is to add a cli hook to conductor to make it possible for users to manually change the status of lost steps when this happens so the workflow can continue?  There'd have to be logging of this of course so the provenance doesn't become unreliable.

maestrowf/interfaces/script/slurmscriptadapter.py
jwhite242: I'll play the format options and see, but it's not clear that restricting state to 2 letters will actually do the same thing since running is only a 1 letter abbreviation.

FrankD412: Sorry -- I was unclear. My question was whether or not we need the `or` clause in there.

FrankD412: Though -- if we fall back to `squeue` if `sacct` doesn't work, then we'd need those there.

FrankD412:
I think it's a good idea to have the backup since it's seemingly not guaranteed that `sacct` will actually be available (as it seems to not be usable on the clusters I'm on).

FrankD412 (approved):

PR #395, opened 2022-05-16
added flux adaptor v0.31.0

bhatiaharsh:
* copied over the v0.26 file to a new one for v0.31.
* added a new "latest" version value.

simply followed the existing style and strategy to expand to a new version. are we going to need a new file for each version?

PR #407, opened 2022-11-26
Introduces additional states for handling new versions of Flux

FrankD412:
This MR introduces state changes that allow Maestro to work with newer versions of Flux that have new state abbreviations for job status. The change is pulled from @Jmast's comment [here](https://github.com/LLNL/maestrowf/pull/395#issuecomment-1277741739).

### DIFF
--- a/maestrowf/interfaces/script/_flux/flux0_26_0.py
+++ b/maestrowf/interfaces/script/_flux/flux0_26_0.py
@@ -2,8 +2,13 @@ import logging
 from math import ceil
 import os
 
-from maestrowf.abstracts.enums import CancelCode, JobStatusCode, State, \
-    StepPriority, SubmissionCode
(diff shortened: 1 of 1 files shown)


FrankD412:
@jwhite -- might it be time to drop some older support for Flux versions and settle on the latest going forward?

jwhite242:
> @jwhite -- might it be time to drop some older support for Flux versions and settle on the latest going forward?

we should definitely make the newest one the default.  should we mark a few of the old ones deprecated first? -> i.e. drop 0.12 and 0.17 in a future release (dunno if we need to wait for 2.0 for this?)

jwhite242 (approved):

FrankD412:
Yeah -- I'm starting to think that might be worthwhile as Flux becomes more mature.

PR #412, opened 2023-01-29
Update versions of packages in poetry.lock for dependabot

FrankD412:
There have been some recent warnings from dependabot about security issues related to `py` and `setuptools`. This MR updates the dependencies so that we are using more recent dependencies. This MR also updates to the most recent `poetry-core` and updates `pytest` with a minimum requirement to remove `py` as a dependency.

*NOTE*: This MR removes Python 3.7 as a minimum version -- it seems that `py` is required for anything lower than 3.8. It seems that Python 3.7 is EOL in June, so we're slightly ahead on deprecating it in Maestro.


doutriaux1:
@FrankD412 do you want to remove 3.7 from the test section as well? Or do we want it for testing?

FrankD412:
Oh it looks like I missed the second version matrix. I'll fix that. Good catch @doutriaux1

pyproject.toml
```diff
@@ -64,6 +64,10 @@ tox-pyenv = "*"
 "conductor" = "maestrowf.conductor:main"
 
 
+
+[tool.poetry.group.dev.dependencies]
+pytest = ">=7.2.0"
```
jwhite242, line 69: this one need to be specifically 7.2 for any reason vs the more permissive 7.x?

jwhite242 (changes requested):
I'd say lets leave 3.7 in here for a while as there are still quite a few users on that version.  Also, is it too early to flip the switch on 3.11 in the matrix?

Additionally, be sure and remove fabric from the dev-dependencies in pyproject.toml as well so poetry doesn't try and pick it up again later.

pyproject.toml
FrankD412: Odd -- it's listed as `'*'` on my end.

FrankD412: Oh -- nevermind this is in a different group.

FrankD412:
> I'd say lets leave 3.7 in here for a while as there are still quite a few users on that version. Also, is it too early to flip the switch on 3.11 in the matrix?
> 
> Additionally, be sure and remove fabric from the dev-dependencies in pyproject.toml as well so poetry doesn't try and pick it up again later.

Alright removed `fabric` and re-add 3.7 -- gonna test adding 3.11 to see how that goes.

jwhite242:
ok, looks like we need one last change: update the sphinx version in the lock file to get the rtd builds passing again.  lock file has it at 3.5, and would appear we need at least > 4.2 to support python >= 3.10.  unsure if there's any reason to pin it to anything earlier than 5.x, but it's not long for the builds anyway so up to you on the version.

FrankD412:
> ok, looks like we need one last change: update the sphinx version in the lock file to get the rtd builds passing again. lock file has it at 3.5, and would appear we need at least > 4.2 to support python >= 3.10. unsure if there's any reason to pin it to anything earlier than 5.x, but it's not long for the builds anyway so up to you on the version.

All fixed!

PR #415, opened 2023-03-08
Patch flux adapter to enable pickling to save study state and update it's resource/uri handling

### LINKED ISSUE #416: Workflow examples with Flux?
Hiya! I'm wondering if you know of any workflow examples that use Maestro and Flux? And ideally has a bit of complexity beyond "hello world!" I am testing Maestro with the Flux Operator, and just had my first view of it via Merlin, which adds a redis/rabbitmq service, and I'm curious to try something without Merlin but still using maestro (to learn better how things work). Thanks for the help!

jwhite242:
Casts flux's JobID objects into string form to enable the dag to be pickled when conductor saves state.  JobID objects are tied to flux's use of cffi and are not picklable themselves.

Base58 string form is used instead of the integer it's derived from so that logging is consistent with the view users have on the commandline with flux reporting job id's in that format.  This requires a separate re-instantiation of JobID objects upon waking to check the job status'.

Additional fixes:

- Improve the flexibility of resource specification in the flux adapter to align with the other adapters (i.e. not requiring nodes be specified).
- Change uri handling so env vars are not the only way to schedule jobs
- Updates to get working on flux only machines that require no bootstrapping or uri
  - Changes job submit call to be non-waitable by default on adapter 0.49 and newer as only owner of instance can do so (which is rarely the case on flux managed machines)
- Adds documentation and recipes for how to use Maestro with flux machines and flux instances running on non-flux scheduled machines
- Pins poetry version to 1.4.2 in CI to maintain support for python 3.7
- Fixes parsing of flux broker version to properly handle non-release versions of flux and logs that version into the flux batch script header comments and the maestro logs.

Resolves #416


vsoch:
@jwhite242 I am using this branch for container builds for the examples I'm working on - do you want any feedback / help / what is the plan for merging this in (and then I can fall back to using the main branch!)

jwhite242:
> @jwhite242 I am using this branch for container builds for the examples I'm working on - do you want any feedback / help / what is the plan for merging this in (and then I can fall back to using the main branch!)

@vsoch Apologies for letting this languish for a bit.   Should be merging this one soon now that the last blocker for submitting jobs so flux managed systems has been removed.   Hopefully the updated docs also help get non-flux experts up and running a little quicker too; the docs you've been adding to new flux docs pages on this have been super helpful there as well!  If you see anything missing in the docs here let me know.

PR #419, opened 2023-05-09
Bugfix/jwhite242/versioninfo

jwhite242:
Update version info to sync up maestro's cli with pyproject.toml for single source of version.


maestrowf/__init__.py
```diff
@@ -41,6 +41,8 @@
 import logging
 import sys
 
+from importlib_metadata import version  # Replace upon deprecation of python 3.7
```
FrankD412, line 44: Happen to know what this gets replaced with once 3.7 is deprecated? Is there an existing solution now?

FrankD412: Oh scratch that -- this is the backport. I see. I know it's a bit uglier but, mind replacing this with the usual
```python
try:
    from importlib.metadata import version
except ImportError as imp_err:
    from importlib_metadata import version
```

I think it makes it clearer that there's a backport and then when 3.7 is EOL'ed we can just remove the except.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ python = ">=3.7.1,<4.0"
 coloredlogs = "*"
 dill = "*"
 filelock = "*"
+importlib_metadata = "*"
```
FrankD412, line 40: This one is out of curiosity, but I think there's a way to conditionally install based on Python version if I'm not mistaken?

FrankD412 (changes requested):

pyproject.toml
jwhite242: There is, but it's already there from some other dependency we have; happy to make it conditional if the dependencies can still override this.

maestrowf/__init__.py
jwhite242: Done.

FrankD412 (approved):

PR #420, opened 2023-06-05
feature/bgunnar5/status-pager

bgunnar5:
Losing my mind with celery atm so I took a break and added the status pager functionality for you. This is my first contribution to Maestro so let me know if there's anything else I need to do that I haven't yet.


FrankD412:
@bgunnar5  -- Thanks for the contribution! I'll review it this week and get back to you. For now, the changes at a glance seem reasonable, I'd just like to test how this behaves locally. Also, it looks like the doc building step in the testing failed so that needs to pass before we can merge this in.

bgunnar5:
Sounds good.

Regarding the docs test failing, it seems to have an import error with the virtualenv package (which shouldn't be a result of anything I modified):
```
ImportError

  cannot import name 'Python2Supports' from 'virtualenv.create.describe' (/home/docs/checkouts/readthedocs.org/user_builds/maestrowf/envs/420/lib/python3.10/site-packages/virtualenv/create/describe.py)

  at ~/checkouts/readthedocs.org/user_builds/maestrowf/envs/420/lib/python3.10/site-packages/virtualenv/create/via_global_ref/builtin/python2/python2.py:6 in <module>
        2│ import json
        3│ import os
        4│ from pathlib import Path
        5│ 
    →   6│ from virtualenv.create.describe import Python2Supports
        7│ from virtualenv.create.via_global_ref.builtin.ref import PathRefToDest
        8│ from virtualenv.info import IS_ZIPAPP
        9│ from virtualenv.util.zipapp import read as read_from_zipapp
       10│
```
@jwhite242 I think this is the same error that Lina is getting in the weave_ci pipeline when she's trying to build the docs. Any idea how to resolve this?

jwhite242:
> 

Maybe..   so the error she had was a different import error, and this one it's not clear what's causing it as the import error message gets injected randomly in poetry's stdout.  I was able to get around it by pinning poetry to 1.4.2 for now, which the readthedocs build doesn't do (1.5.x fails for me with the same error).   So try changing this line to pin poetry to that version: https://github.com/LLNL/maestrowf/blob/f5fcad59a697a389359b00c8faf1576050924404/.readthedocs.yaml#L9

jwhite242:
One question on the pager/theme issues you'd found: is there an exception that gets thrown, or is it just a terminal emulator error that we can't do anything about in here?  Meaning, there's no sort of try/except block we can do to auto-rollback the use styles flag in the pager here?

bgunnar5:
I haven't seen any errors thrown when using pager and theme together. If the pager doesn't accept customization it'll just display something like this in the terminal window:
image

I don't believe there's a way to try/except catch this. I did add a message in the `maestro status --help` page for the `--disable-theme` description talking about how setting MANPAGER or PAGER to be "less -r" may be the fix for this. Maybe I could add a little printed message to be displayed on each status run that says this instead so users don't have to go searching for it?

doutriaux1:
@bgunnar5 maybe here? https://stackoverflow.com/questions/7445658/how-to-detect-if-the-console-does-support-ansi-escape-codes-in-python

bgunnar5:
this specific issue is weird @doutriaux1 because if you don't use the pager functionality then the ANSI colors work just fine, so I'm not quite sure if this relates to stdout/stderr or if this is something different entirely

jwhite242:
> 

Ok, thanks for checking.  Just want to make sure we catch things when we can.

jwhite242:
So I think the only thing really missing now is adding these options to the docs.  The cli args for sure need to get put in here: https://github.com/LLNL/maestrowf/blob/f5fcad59a697a389359b00c8faf1576050924404/docs/Maestro/cli.md?plain=1

But maybe also one of the admonitions (https://squidfunk.github.io/mkdocs-material/reference/admonitions/) would be good in the montoring section of the user guide (https://github.com/LLNL/maestrowf/blob/f5fcad59a697a389359b00c8faf1576050924404/docs/Maestro/monitoring.md?plain=1)  to mention what to do if things look garbled.   Not sure which one's most appropriate here, Tip, Note, Info,...  or maybe even Warning?

bgunnar5:
Good call, can't believe I forgot docs! I added a note admonition in the monitoring section and two sections for the theme and pager additions that you should easily be able to expand on in the future if you'd like.

docs/Maestro/monitoring.md
```diff
@@ -75,7 +95,7 @@ $ maestro status ./tests/lulesh --layout narrow
 A snippet of the narrow layout for the above study is shown below.  These layouts are computed by status command, so you can alternate between them in the same study without issue:
 
 <figure markdown>
-  ![Flat Layout Lulesh Cancelled](../assets/images/examples/study_monitoring/narrow_layout_lulesh_cancelled.png)
+  ![Narrow Layout Lulesh Cancelled](../assets/images/examples/study_monitoring/narrow_layout_lulesh_cancelled.png)
```
FrankD412, line 98: I think this image might be missing? -- at least it doesn't show up in the list of changed files on my end.

bgunnar5:
No worries! Jeremy and I were discussing these changes a bit before I even made this branch so he's pretty up-to-date on everything I've been doing here 😄

docs/Maestro/monitoring.md
bgunnar5: I didn't add the image here, I just modified the label since I noticed it was incorrect

PR #424, opened 2023-07-18
Add execution order controls

jwhite242:
Add machinery for controlling execution order/priority of study steps

* Adds weights to the graph to enable selecting between depth first and breadth first (current production mode) execution order.
* Adds new execution block for exposing controls of orders and future hooks for using various step metadata for controlling step weights
* Changes internal machinery to use a PriorityQueue to store the ready steps

### DIFF
--- a/docs/Maestro/specification.md
+++ b/docs/Maestro/specification.md
@@ -378,253 +378,257 @@ execution:
     step_order: depth-first
 ```
 
-These execution orders can be shown with the sample hello-bye world specification as shown in the tutorials:
+!!! warning
 
-``` yaml
-description:
-    name: hello_bye_parameterized_funnel
-    description: A study that says hello and bye to multiple people, and a final good bye to all.
-
-execution:
-    step_order: depth-first
-
-env:
-    variables:
-        OUTPUT_PATH: ./samples/hello_bye_parameterized_funnel
-    labels:
-        HELLO_FORMAT: $(GREETING)_$(NAME).txt
-        BYE_FORMAT: $(FAREWELL)_$(NAME).txt
-
-study:
-    - name: say-hello
-      description: Say hello to someone!
-      run:
-          cmd: |
-            echo "$(GREETING), $(NAME)!" > $(HELLO_FORMAT)
-
-    - name: say-bye
-      description: Say bye to someone!
-      run:
-          cmd: |
-            echo "$(FAREWELL), $(NAME)!" > $(BYE_FORMAT)
-          depends: [say-hello]
-
-    - name: bye-all
-      description: Say bye to everyone!
-      run:
-          cmd: |
-            echo "Good-bye, World!" > good_bye_all.txt
-          depends: [say-bye_*]
-
-global.parameters:
-    NAME:
-        values: [Pam, Jim, Michael, Dwight]
-        label: NAME.%%
-    GREETING:
-        values: [Hello, Ciao, Hey, Hi]
-        label: GREETING.%%
-    FAREWELL:
-        values: [Goodbye, Farewell, So long, See you later]
-        label: FAREWELL.%%
-```
-
-!!! note
-
-    This execution can interact with the [throttle](cli.md#run) setting,
-    resulting in a mix of breadth/depth-first ordering.  This is due to the step
-    dependencies limiting what can be put into the queue: ``say-bye`` steps
+    For [throttle](cli.md#run) > 1 the ``depth-first`` option will not behave as a pure
+    depth-first traversal, but rather a mix of breadth/depth-first ordering.  The step
+    dependencies limit what can be put into the queue: ``say-bye`` steps
     cannot enter the queue until their dependencies are met, e.g. the
     corresponding ``say-hello`` steps.  Thus for throttle settings > 1, you may
-    have multiple ``say-hello`` steps run before any ``say-bye`` steps.  For the
-    two modes diagrammed below we will assume throttle = 1 to make things more
+    have multiple ``say-hello`` steps run before any ``say-bye`` steps.  The
+    two modes diagrammed below assume throttle = 1 to make things more
     clear.
-
-
-Breadth first, order of excution marked by colors with lighter colors executing first:
-
-``` mermaid
-flowchart TD;
-    A(study root) --> COMBO1;
-    subgraph COMBO1 [Combo #1]
-      subgraph say_hello1 [say-hello]
-        B(Hello, Pam)
-
-      end
-      subgraph say_bye1 [say-bye]
-        C(Goodbye, Pam)
-        
-      end
-      say_hello1 --> say_bye1
-    end
-    A --> COMBO2
-    subgraph COMBO2 [Combo #2]
-      direction TB
-      subgraph say_hello2 [say-hello]
-        D(Ciao, Jim)
-        
-      end
-      subgraph say_bye2 [say-bye]
-        E(Farewell, Jim)
-        
-      end
-      say_hello2 --> say_bye2
-    end
-    A --> COMBO3
-    subgraph COMBO3 [Combo #3]
-      subgraph say_hello3 [say-hello]
-        F(Hey, Michael)
-        
-      end
-      subgraph say_bye3 [say-bye]
-        G(So long, Michael)
-        
-      end
-      say_hello3 --> say_bye3
-    end
-    A --> COMBO4
-    subgraph COMBO4 [Combo #4]
-      subgraph say_hello4 [say-hello]
-        H(Hi, Dwight)
-        
-      end
-      subgraph say_bye4 [say-bye]
-        I(See you later, Dwight)
-        
-      end
-      say_hello4 --> say_bye4;
-    end
-    
-    COMBO1 --> J{{bye-all}}
-    COMBO2 --> J{{bye-all}}
-    COMBO3 --> J{{bye-all}}
-    COMBO4 --> J{{bye-all}}
     
     
-    style B fill:#fff7f3,color:#000000;
-    style C fill:#f768a1,color:#000000;
-    style D fill:#fde0dd,color:#000000;
-    style E fill:#dd3497,color:#FFFFFF;
-    style F fill:#fcc5c0,color:#000000;
-    style G fill:#ae017e,color:#FFFFFF;
-    style H fill:#fa9fb5,color:#000000;
-    style I fill:#7a0177,color:#FFFFFF;
-    style J fill:#49006a,color:#FFFFFF;
+These execution orders can be shown with the sample hello-bye world specification as shown in the tutorials:
 
-    subgraph LEGEND [Legend: Execution order]
-      direction LR
-      BL[1]---> DL[2]
-      DL ---> FL[3]
-      FL ---> HL[4]
-      HL ---> CL[5]
-      CL ---> EL[6]
-      EL ---> GL[7]
-      GL ---> IL[8]
-      IL ---> JL[9]
-      style BL fill:#fff7f3,color:#000000;
-      style CL fill:#f768a1,color:#000000;
-      style DL fill:#fde0dd,color:#000000;
-      style EL fill:#dd3497,color:#FFFFFF;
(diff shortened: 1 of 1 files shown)


docs/Maestro/specification.md
```diff
@@ -363,6 +364,15 @@ The `batch` block is an optional block that enables specification of HPC schedul
...
 <br/>
 
+## Execution: `execution`
+----
+
+The `execution` block is an optional block that enables control of the study execution.  The current version allows selection between depth first and breadth first order of execution of the study steps.  Breadth first order is the default mode.
+
+|  **Key**      |  **Required**  | **Type** | **Description** |
+|    :-         |      :-:       |    :-:   |       :-        |
+|  `step_order` |      No        |   str    | Type of scheduler managing execution.  One of: {`depth-first`, `breadth-first`}. Default: `depth-first`. |
```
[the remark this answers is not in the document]
jwhite242: yeah, that's definitely a typo.  should be fixed now

maestrowf/datastructures/execution/execution.py
```diff
@@ -0,0 +1,91 @@
...
+        if self.step_order == 'breadth-first':
+            self.step_prioritization_factory.register_priority_expr(
+                'step_order',
+                step_weight_priority_bf
+            )
+        else:
+            self.step_prioritization_factory.register_priority_expr(
+                'step_order',
+                step_weight_priority_df
+            )
```
bgunnar5, line 88: I'm not entirely certain on your future plans for this section of code, but if you're planning on adding more additions to this if/else chain of the same form then maybe it would be better in the long run if you did something like so:
```
priority_expr_mapper = {
    "breadth-first": {
        "name": "step-order",
        "func": step_weight_priority_bf
    },
    "depth-first": {
        "name": "step-order",
        "func": step_weight_priority_df
    }
}

self.step_prioritization_factory.register_priority_expr(
    priority_expr_mapper[self.step_order]["name"],
    priority_expr_mapper[self.step_order]["func"]
)
```
Then any changes to the `register_priority_expr()` method would only have to be modified in one function call rather than multiple.

bgunnar5 (commented):
I like this addition a lot. Two of my comments here are just suggestions for cleaning up code/planning for the future but if you want to ignore them I'm not going to be offended at all.

tests/specification/test_specs/prioritized_lulesh_sample1_unix.yml
```diff
@@ -0,0 +1,93 @@
+description:
```
jsemler, line 1: It would be helpful to have an example in the samples directory as well.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -906,12 +920,32 @@ def execute_ready_steps(self):
             # Now, we need to take the min of the length of the queue and the
             # computed number of slots. We could have free slots, but have less
             # in the queue.
-            _available = min(_available, len(self.ready_steps))
+            # _available = min(_available, len(self.ready_steps))  # Kind of inconsistent if it's affected by queue size sometimes
```
FrankD412, line 923: I'm not 100% sure what you mean by inconsistent here. This prevents the need for the conditional below (at least that's what it's intent was when it was originally coded. It means not checking for the `-1` value explicitly also.

maestrowf/datastructures/core/executiongraph.py
```diff
@@ -906,12 +920,32 @@ def execute_ready_steps(self):
...
+        # for item in que_steps:
+        #     pprint(item)
+
+        
+        # for step_order, step in enumerate(self.ready_steps):
+        #     pprint(f"{step_order}: {step[0]}")
+        new_steps_running = 0
+        LOGGER.debug("Available slots: %d, Ready steps is empty %s",
+                     _available, self.ready_steps.empty() == True)
+        while not self.ready_steps.empty() and (new_steps_running < _available or _available < 0):
```
FrankD412, line 942: This changes the original intent of the throttle that was added. The `--throttle` option meant that Maestro would not exceed that number of jobs in total. This form of throttle says that we are allowed to launch that number of new steps instead of being a global limit to my understanding. Am I incorrect there?

Also, this conditional is a little convoluted -- since `available` could be `-1` it adds the extra conditional.

maestrowf/datastructures/core/study.py
```diff
@@ -127,6 +130,48 @@ def real_name(self):
...
+        """
+        Less than operator for the StudyStep class.
+
+        :param other: Object to compare self to.
+        :returns: True if weight is less than other's weight
+        """
+        if self.weight < other.weight:
+            return True
+
+        return False
```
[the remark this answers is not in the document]
FrankD412: To add onto this, I think you're only required to implement `__lt__` and `__eq__` since you can build the other comparison methods from them. But I'm only 85% sure from memory 😅

maestrowf/datastructures/core/study.py
```diff
@@ -193,8 +255,8 @@ class Study(DAG, PickleInterface):
       using the same environment.
     """
 
-    def __init__(self, name, description,
-                 studyenv=None, parameters=None, steps=None, out_path="./"):
+    def __init__(self, name, description, studyenv=None,
+                 parameters=None, steps=None, out_path="./"):
```
FrankD412, line 259: Since you're passing the study environment here, the output path might be redundant. Could be an opportunity to clean up some parameters here.

maestrowf/datastructures/dag.py
```diff
@@ -127,6 +127,35 @@ def remove_edge(self, src, dest):
         logging.debug("Removing edge (%s, %s).", src, dest)
         self.adjacency_table[src].remove(dest)
 
+    def __iter__(self):
+        """
+        Iterator over the graph
```
FrankD412, line 132: This docstring could be clearer. It also looks like this is a DFS search. It might be worth mentioning that in the docstring or making it configurable somehow.

maestrowf/datastructures/execution/execution.py
FrankD412: Another option here is to have some protocoled classes that implement the `__call__` method -- then you can just point to the class you want it to use and have them all be callable.

maestrowf/datastructures/core/study.py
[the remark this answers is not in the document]
jwhite242: I'll throw another wrench into this and point out this implementation is probably not what we want despite it working for the current feature.  I.e. step weights can't be the only basis of comparison.  The `__eq__` gets used for doing step equality, which means we can't use `__lt__` to construct that if we stick with this `__lt__` that only cares about the step weights.    Had kind of punted on what other things to care about in these during the initial implementation, but should definitely be addressed now

maestrowf/datastructures/execution/execution.py
jwhite242: Yeah, both good ideas, and I think the end game is some mix of the two to get all the work out of the constructor.   Haven't quite fleshed out the design for the implementation, but idea with this is to eventually enable users to build expressions from things like step resource keys (procs, walltime, combinations of them), parameters, maybe even step names and layer them on to this priority.  Priorities then being the sequential eval of all of those expressions (step order being the only one that's always there) and exploiting the sorting mechanism of tuples to give a lot of fine grained control over the orders. 

Of course, if there are other ideas on things we might want to use to control this ordering, let me know!

maestrowf/datastructures/core/executiongraph.py
jwhite242: yeah, this part needs another data structure to really work right again: queue's size method is only approximate, so isn't reliable for this like the previous implementation was.   will need a running steps tracker alongside this to make this throttling less convoluted

FrankD412:
@jwhite242 -- Is this a dead end at this point? We're assessing Maestro and it's looking like DFS would be good on our end too. Wondering just in case we need to revisit.

jwhite242:
> @jwhite242 -- Is this a dead end at this point? We're assessing Maestro and it's looking like DFS would be good on our end too. Wondering just in case we need to revisit.

No, i just got derailed by other things for a bit.  Am ramping back up on this now.  Having the more general expression based priorities are going to be pretty helpful -> major use case here being getting big/long running variants of steps running sooner, allowing smaller ones to be churned through within the throttle limit alongside it for improved throughput.

Thinking more on the protocol question..   I'm on the fence on whether we shouldn't just use abstract base classes and tie info to these things; i.e. per step overrides of expressions.   But will play with both and see how they feel

maestrowf/datastructures/core/executiongraph.py
jwhite242: Problem with this is PriorityQueue's are like standard Queue's and do not implement a length like the dequeue that used to live here.  The size is also only approximate (unsure if this is really only the case in multithreading cases or not)

maestrowf/datastructures/core/executiongraph.py
jwhite242: so, this actually does work as intended, no need to do my edit.  The controls come from `_available` getting set using throttle and the number of running steps a little above this loop.  So if `in_progress` is already = throttle this conditional will not let any new jobs through.  

just tested it on one of the hello world specs and verified it would properly limit to 1 job at a time (throttle setting) with the slurm adapter

Will add a comment to explain the variable.

maestrowf/datastructures/core/executiongraph.py
jwhite242: Also this whole block can now be removed as we don't need this check -> the if queue.empty() test on the while loop accomplishes the same behavior and exits when there are no more steps are available

jwhite242:
@FrankD412, @bgunnar5, @jsemler 
Think this is finally ready for another pass/real review.  An interesting question left (beyond any implementation issues/comments) is what to do about the spec.  I refactored it to be a list so it's more clearly ordered for users, but maybe it'd make sense to contian this list in a subkey (`priority_expressions` or something) instead of at the root of the execution block?  Don't have any other things in mind for this block yet, but thinking the key would be more future proof in case we do think of something.  (i know docs are slightly out of sync, pending this subkey/not question)

jwhite242:
actually, just had another thought that might fit nicer, expanding it and making the value more of a 'oneOf' type, so either value or expression, making it more clear that there's two types and avoiding having to do greedy parsing on things to figure it out on our end

```
execution:
  priority:
    - name:
      description: # optional, but encouraged... can make built-ins dump the code's internal description in the reserialized spec
      value: # use this for built-ins with string keys to select (e.g. current 'step-order')
    - name:
      description:
      expression: # use this for the eventual string based expression compilation
```

jwhite242:
> actually, just had another thought that might fit nicer, expanding it and making the value more of a 'oneOf' type, so either value or expression, making it more clear that there's two types and avoiding having to do greedy parsing on things to figure it out on our end
> 
> ```
> execution:
>   priority:
>     - name:
>       description: # optional, but encouraged... can make built-ins dump the code's internal description in the reserialized spec
>       value: # use this for built-ins with string keys to select (e.g. current 'step-order')
>     - name:
>       description:
>       expression: # use this for the eventual string based expression compilation
> ```

Continued tweaking/iteration with a mind toward this being amenable to a mix of built-in/user things (think plugins for reusable expressions checkable via the dependencies machinery)

```
```

tests/study/test_study.py
```diff
@@ -0,0 +1,142 @@
...
+    test_study.configure_study()
+    test_study.setup_environment()
+    test_out_path, test_exec_graph = test_study.stage()
+
+    base_step_names = list(expected_step_weights.keys())
+    for parameterized_step_name, parameterized_step in test_exec_graph:
+        pprint(f"Parameterized step name: {parameterized_step_name}, step: {parameterized_step}")
+        if not parameterized_step:  # catch source, which is always none
+            continue
+        # NOTE: really can't get the base step name anymore after staging?
```
bgunnar5, line 71: I ran into this same problem when developing Merlin's status commands so if we could implement an easy way to access this post-staging that would be _amazing_

molohov:
Any chance of moving this along? I would love to have DFS implemented for our use case.

PR #428, opened 2023-10-04
Fix flux job cancellation

jwhite242:
Update 0.49 and 0.26 adapters cancellation to deal with recent update to submit that converts from flux JobID to native types.

Update general cancellation behaviors:
- Check for in progress steps before declaring cancelled successfully to delay until actual final states can be serialized
- Update cancel logic to mirror failure logic: mark all steps downstream of a cancelled step to also be cancelled


FrankD412 (approved):

jwhite242:
@FrankD412 You may want to revoke/revisit that review as testing led me to chase down some other oddities and make more significant behavioral changes to step cancellations in general

Issue #441, opened 2024-05-02
jobs stuck in PENDING state but they've actually completed (conductor process is still running)


BenWibking:
I've encountered the confusing situation of having the conductor still running as a background process, my jobs have all completed successfully, but running `maestro status` lists the jobs that were run via SLURM as still "PENDING".

Is there any way to figure out what has gone wrong, or otherwise reset the conductor process?

Running `ps aus | grep $USER`, I see:
```
login4.stampede3(626)$ ps aux | grep $USER
bwibking 1549018  0.0  0.0  20820 11936 ?        Ss   Apr30   0:09 /usr/lib/systemd/systemd --user
bwibking 1549019  0.0  0.0 202568  6848 ?        S    Apr30   0:00 (sd-pam)
bwibking 1725176  0.0  0.0   8436  3692 ?        S    Apr30   0:00 /bin/sh -c nohup conductor -t 60 -d 2 /scratch/02661/bwibking/precipitator-paper/outputs/medres_compressive_20240430-160408 > /scratch/02661/bwibking/precipitator-paper/outputs/medres_compressive_20240430-160408/medres_compressive.txt 2>&1
bwibking 1725177  0.0  0.0 325552 67876 ?        S    Apr30   0:32 /scratch/projects/compilers/intel24.0/oneapi/intelpython/python3.9/bin/python3.9 /home1/02661/bwibking/.local/bin/conductor -t 60 -d 2 /scratch/02661/bwibking/precipitator-paper/outputs/medres_compressive_20240430-160408
root     3152422  0.0  0.0  39932 11976 ?        Ss   13:31   0:00 sshd: bwibking [priv]
bwibking 3152677  0.0  0.0  40240  7472 ?        S    13:31   0:00 sshd: bwibking@pts/122
bwibking 3152678  0.0  0.0  17920  6116 pts/122  Ss   13:31   0:00 -bash
root     3163591  0.0  0.0  39932 12044 ?        Ss   13:43   0:00 sshd: bwibking [priv]
bwibking 3163629  0.0  0.0  40076  7236 ?        S    13:43   0:00 sshd: bwibking@notty
root     3168472  0.0  0.0  39932 12020 ?        Ss   13:48   0:00 sshd: bwibking [priv]
bwibking 3168505  0.0  0.0  40080  7208 ?        S    13:48   0:00 sshd: bwibking@notty
bwibking 3174162  0.0  0.0  20408  4220 pts/122  R+   13:54   0:00 ps aux
[...]

PR #443, opened 2024-05-08
make parsing jobid from sbatch output more robust

BenWibking:
This adds `--parsable` to the `sbatch` options when submitting a SLURM job. Then, when looking for the jobid returned by SLURM, it only examines the last line of output, which should conform to the description given in the documentation (https://slurm.schedmd.com/sbatch.html#OPT_parsable).

Fixes https://github.com/LLNL/maestrowf/issues/441.

### DIFF
--- a/maestrowf/interfaces/script/slurmscriptadapter.py
+++ b/maestrowf/interfaces/script/slurmscriptadapter.py
@@ -217,6 +217,8 @@ class SlurmScriptAdapter(SchedulerScriptAdapter):
         """
         # Leading command is 'sbatch'
         cmd = ["sbatch"]
+        # add "--parsable": https://slurm.schedmd.com/sbatch.html#OPT_parsable
+        cmd += ["--parsable"]
         # Check and see if we should be submitting into a reservation.
         if "reservation" in self._batch:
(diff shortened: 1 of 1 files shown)


jwhite242 (commented):
Ok, so only comment i think i have here is a slight change to the regex to ensure we can't match numbers that are part of other words  (i.e. it finding the '3' in 'Stampede3').  So how about tweaking the regex to something like the following to ensure it can only find numbers attached to slurm's nominal sbatch output.  Something like the following, with or without the named capture groups:

`(?:Submitted batch job \b(?P<jobid>\d+)\b)`

or maybe don't bother capturing the first part but require it to match:

`Submitted batch job \b(?P<jobid>\d+)\b`

Likely not worth the hassle to get lookahead/lookbehind constructs to filter out everything but the number as I think one of them isn't even supported in python?

jwhite242:
Oh, and one more comment.  Having the full string in the match means we shouldn't have to assume it's the last line either, so just using the regex.search iterator should find a match on any line; should be safer in case those plugins also allow output after the jobid (would hope not, but easy enough to guard against that).

PR #459, opened 2025-08-15
Extra Flux Options

jwhite242:
Expands available options to pass through to flux, replacing the `args` dictionary in the batch block which only passed through shell options (--setopt/-o) to the launcher (flux run).  Adds two dictionaries to the batch block:

* `allocation_args`: enable attaching arbitrary numbers of options to the allocation (jobspec) corresponding to the cli equivalents: `-o/--setopt`, `-S/--setattr`, and `--conf`
* `launcher_args`: replaces `args` for attaching to launcher, but enables passing through any option prefix, not just `-o/--setopt`

Deprecates `args`, but leaves it intact as the only option for older flux adapters (< 0.49) pending future removal.

Adds flux directives with all allocation options to serialized step scripts to enable reproducibility and direct cli submission by users to match capability of other script adapters.

Exposes new exclusive syntax for step keys to enable separate allocation/launcher control: only flux uses the launcher option, other schedulers to follow.

Adds script writing and job submission tests for both flux and slurm.

Updates info directive in flux scripts from INFO to MAESTRO-INFO for more clearly showing source of info.

### DIFF
--- a/docs/Maestro/scheduling.md
+++ b/docs/Maestro/scheduling.md
@@ -23,9 +23,13 @@ steps:
 | `procs`       |      No        |   int    | Optional number of tasks in batch allocations: note this is also a per step key |
 | `flux_uri`    |      Yes*      |   str    | URI of the Flux instance to schedule jobs to. * Only used with `type`=`flux`. NOTE: It is recommended to rely on environment variables instead, as URIs are very ephemeral and may change frequently.|
 | `version`     |      No        |   str    | Optional version of flux scheduler; for accommodating api changes |
-| `args`        |      No        |   dict   | Optional additional args to pass to scheduler; keys are arg names, values are arg values |
+| :warning: `args`        |      No        |   dict   | Optional additional args to pass to scheduler; keys are arg names, values are arg values |
+| `allocation_args` | No  | dict | Optional scheduler specific options/flags to add to allocations :material-information-slab-circle: flux only in :material-tag:`1.1.12` |
+| `launcher_args` | No  | dict | Optional scheduler specific options/flags to add to launcher commands :material-information-slab-circle: flux only in :material-tag:`1.1.12` |
 
+!!! warning "`args` deprecated"
 
+    `args` has been marked deprecated in :material-tag: `1.1.12` in favor of the more flexible `allocation_args` and `launcher_args`
 
 The information in this block is used to populate the step specific batch scripts with the appropriate
 header comment blocks (e.g. '#SBATCH --partition' for slurm).  Additional keys such as step specific
@@ -39,6 +43,26 @@ run locally unless at least the ``nodes`` or ``procs`` key in the step is popula
     2.  Flux brokers may not always have named queues (nested allocations).
 	
 	See [queues and banks](how_to_guides/running_with_flux.md#queues-and-banks) section in the how-to guide on running with flux for more discussion.
+
+### Extra Arguments
+---
+
+There are new groups in the batch block in :material-tag:`1.1.12` that facilitate adding custom options to both allocations and the `$(LAUNCHER)` invocations independently.  These are grouped into two dictionaries in the batch block which are meant to enable passing in options that Maestro cannot abstract across schedulers more generally:
+
+* `allocation_args` for the allocation target (batch directives such as `#Flux: --setopt=foo=bar`)
+* `launcher_args` for the `$(LAUNCHER)` target (`flux run --setopt=foo=bar`)
+
+These are ~structured mappings which are designed to mimic cli usage for intuitive mapping from raw scheduler usage to Maestro.  Each of these dictionaries' keys correspond to a scheduler specific CLI argument/option or flag.  The serialization rules are as follows, with specific examples here shown for the initial implementation in the flux adapter (other schedulers will yield prefix/separator rules specific to their implementation):
+
+|  **Key Type**   | **CLI Prefix** | **Separator** | **Example YAML**    | **Example CLI Output** |
+|    :-           | :-:            |     :-:       |    :-:              |       :-               |
+|  Single letter  |  `-`           | space (`" "`) | `o: {bar: 42}`      | `-o bar=42`            |
+|  Multi-letter   | `--`           |  `=`          | `setopt: {foo: bar} | `--setopt=foo=bar`     |
+|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: } | `--setopt=foobar`      |
+|  Boolean flag w/o key   | as above       | as above      | `exclusive: ` | `--exclusive`      |
+
+Note in the boolean flag strategies, a space is required after the `:` after `foobar: ` or `exclusive`, otherwise yaml will fail to parse and assign the Null value used to tag a key as a boolean flag.  See [flux](#Flux) for special considerations fo the `allocation_args`
+
 
 ## LAUNCHER Token
 ---
@@ -187,6 +211,64 @@ See the [flux framework](https://flux-framework.readthedocs.io/en/latest/index.h
     The Flux scheduler itself and Maestro's flux adapter are still in a state of flux and may go through breaking changes more frequently than the Slurm and LSF scheduler adapters.
    
 
+### Extra Flux Args
+----
+
+As of :material-tag:`1.1.12`, the flux adapter takes advantage of new argument pass through for scheduler options that Maestro cannot abstract away.  This is done via `allocation_args` and `launcher_args` in the batch block, which expand upon the previous `args` input which only applied to `$(LAUNCHER)`.  There are some caveat's here due to the way Maestro talks to flux.  The current flux adapters all use the python api's from Flux to build the batch jobs, with the serialized batch script being serialized separately instead of submitted directly as with the other schedulers.  A consequence of this is the `allocation_args` map to specific call points on that python api, and thus the option pass through is not quite arbitrary.  There are 4 currently supported options for allocations which cover a majority of usecases (open an issue and let us know if one you need isn't covered!):
+
+* shell options: `-o/--setopt` prefixed arguments
+* attributes: `-S/--setattr` prefixed arguments
+* conf: `--conf` prefixed arguments
+* exclusive flags: `-x, --exclusive` are used to set defaults, with step exclusive keys overriding
+
+!!! warning
+
+    All other flags will be allowed in `allocation_args`, but they will essentially be ignored when serializing the step scripts and submitting jobs
+	
+The `launcher_args` (`$(LAUNCHER)`) will pass through anything as it is a string generator just like other script adapters.  :warning: These are not validated!  Passing arguments that flux doesn't know what to do with may result in errors.
+
+#### Example Batch Block
+---
+
+``` yaml
+batch:
+  type: flux
+  host: machineA
+  bank: guests
+  queue: debug
+  allocation_args:
+    setopt:
+      foo: bar
+    o:
+      bar: 42
+    setattr:
+      foobar: "whoops"
+    conf:
+      resource.rediscover: "true"  # Use string "true" for Flux compatibility, not "True" or bool True
+  launcher_args:
+    setopt:
+      optiona:   # Boolean flag, no value needed
+```
+
+#### Example Batch Script
+---
+Assuming the step has keys `{procs: 1, nodes: 1, cores per task: 1, walltime: "5:00"}`:
+
+``` console
+#flux: -q debug
+#flux: --bank=guests
+#flux: -t 300s
+#flux: --setopt=foo=bar
+#flux: -o bar=42
+#flux: --setattr=foobar=whoops
+#flux: --conf=resource.rediscover=true
+
+flux run -n 1 -N 1 -c 1 --setopt=optiona  myapplication
+```
+
+!!! note
+
+    Using flux directives here to illustrate even though python api is used.  These directives will be in the step scripts, retaining repeatability/record of what was submitted and viewable with the dry run feature
    
 ## LSF: a Tale of Two Launchers
 ----
--- a/docs/Maestro/scheduling.md
+++ b/docs/Maestro/scheduling.md
@@ -52,16 +52,18 @@ There are new groups in the batch block in :material-tag:`1.1.12` that facilitat
 * `allocation_args` for the allocation target (batch directives such as `#Flux: --setopt=foo=bar`)
 * `launcher_args` for the `$(LAUNCHER)` target (`flux run --setopt=foo=bar`)
 
-These are ~structured mappings which are designed to mimic cli usage for intuitive mapping from raw scheduler usage to Maestro.  Each of these dictionaries' keys correspond to a scheduler specific CLI argument/option or flag.  The serialization rules are as follows, with specific examples here shown for the initial implementation in the flux adapter (other schedulers will yield prefix/separator rules specific to their implementation):
+These are ~structured mappings which are designed to mimic cli and batch script directive syntax for intuitive mapping from raw scheduler usage to Maestro.  Each of these dictionaries' keys correspond to a scheduler specific CLI argument/option or flag.  The serialization rules are as follows, with specific examples here shown for the initial implementation in the flux adapter (other schedulers will yield prefix/separator rules specific to their implementation):
 
-|  **Key Type**   | **CLI Prefix** | **Separator** | **Example YAML**    | **Example CLI Output** |
-|    :-           | :-:            |     :-:       |    :-:              |       :-               |
-|  Single letter  |  `-`           | space (`" "`) | `o: {bar: 42}`      | `-o bar=42`            |
-|  Multi-letter   | `--`           |  `=`          | `setopt: {foo: bar} | `--setopt=foo=bar`     |
-|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: } | `--setopt=foobar`      |
-|  Boolean flag w/o key   | as above       | as above      | `exclusive: ` | `--exclusive`      |
+|  **Key Type**         | **Prefix** | **Separator** | **Example YAML**     | **Example CLI Input/Directive** |
+|    :-                 | :-            | :-       |    :-                |       :-               |
+|  Single letter        |  `-`           | `" "` (space) | `o: {bar: 42}`       | `-o bar=42`            |
+|  Multi-letter         | `--`           |  `=`          | `setopt: {foo: bar}` | `--setopt=foo=bar`     |
+|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: }` | `--setopt=foobar`      |
+|  Boolean flag w/o key | as above       | as above      | `exclusive: `        | `--exclusive`      |
 
-Note in the boolean flag strategies, a space is required after the `:` after `foobar: ` or `exclusive`, otherwise yaml will fail to parse and assign the Null value used to tag a key as a boolean flag.  See [flux](#Flux) for special considerations fo the `allocation_args`
+!!! note "Boolean/Flag type arguments"
+
+    In the boolean flag strategies, a space is required after the `:` after `foobar: ` or `exclusive`, otherwise yaml will fail to parse and assign the Null value used to tag a key as a boolean flag.  See [flux](#Flux) for special considerations for the `allocation_args`. 
 
 
 ## LAUNCHER Token
@@ -259,7 +261,7 @@ Assuming the step has keys `{procs: 1, nodes: 1, cores per task: 1, walltime: "5
 #flux: --bank=guests
 #flux: -t 300s
 #flux: --setopt=foo=bar
-#flux: -o bar=42
+#flux: --setopt=bar=42
 #flux: --setattr=foobar=whoops
 #flux: --conf=resource.rediscover=true
 
@@ -268,7 +270,7 @@ flux run -n 1 -N 1 -c 1 --setopt=optiona  myapplication
 
 !!! note
 
-    Using flux directives here to illustrate even though python api is used.  These directives will be in the step scripts, retaining repeatability/record of what was submitted and viewable with the dry run feature
+    Using flux directives here to illustrate even though python api is used.  These directives will be in the step scripts, retaining repeatability/record of what was submitted and viewable with the dry run feature.  The batch/allocation arguments are normalized to the long form (`--setattr` instead of `-S`) and will show up that way in the serialized batch scripts.
    
 ## LSF: a Tale of Two Launchers
 ----
--- a/docs/Maestro/scheduling.md
+++ b/docs/Maestro/scheduling.md
@@ -56,14 +56,14 @@ These are ~structured mappings which are designed to mimic cli and batch script 
 
 |  **Key Type**         | **Prefix** | **Separator** | **Example YAML**     | **Example CLI Input/Directive** |
 |    :-                 | :-            | :-       |    :-                |       :-               |
-|  Single letter        |  `-`           | `" "` (space) | `o: {bar: 42}`       | `-o bar=42`            |
-|  Multi-letter         | `--`           |  `=`          | `setopt: {foo: bar}` | `--setopt=foo=bar`     |
-|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: }` | `--setopt=foobar`      |
-|  Boolean flag w/o key | as above       | as above      | `exclusive: `        | `--exclusive`      |
+|  Single letter        |  `-`           | `" "` (space) | <pre><code><span>o:</span></br><span>  bar: 42</span></code></pre>       | `-o bar=42`            |
+|  Multi-letter         | `--`           |  `=`          | <pre><code><span>setopt:</span></br><span>  foo: bar</span></code></pre> | `--setopt=foo=bar`     |
+|  Boolean flag w/key   | as above       | as above      | <pre><code><span>setopt:</span></br><span>  foobar: #</span></code></pre> | `--setopt=foobar`      |
+|  Boolean flag w/o key | as above       | as above      | `exclusive: #`        | `--exclusive`      |
 
 !!! note "Boolean/Flag type arguments"
 
-    In the boolean flag strategies, a space is required after the `:` after `foobar: ` or `exclusive`, otherwise yaml will fail to parse and assign the Null value used to tag a key as a boolean flag.  See [flux](#Flux) for special considerations for the `allocation_args`. 
+    In the boolean flag strategies, a space is required after the `:` after `foobar: ` or `exclusive`, otherwise yaml will fail to parse and assign the Null value used to tag a key as a boolean flag.  See [flux](#Flux) for special considerations for the `allocation_args`.  See the above examples where a '#' is added after a space to ensure there is such a space after the `:`.
 
 
 ## LAUNCHER Token
@@ -216,7 +216,7 @@ See the [flux framework](https://flux-framework.readthedocs.io/en/latest/index.h
 ### Extra Flux Args
 ----
 
-As of :material-tag:`1.1.12`, the flux adapter takes advantage of new argument pass through for scheduler options that Maestro cannot abstract away.  This is done via `allocation_args` and `launcher_args` in the batch block, which expand upon the previous `args` input which only applied to `$(LAUNCHER)`.  There are some caveat's here due to the way Maestro talks to flux.  The current flux adapters all use the python api's from Flux to build the batch jobs, with the serialized batch script being serialized separately instead of submitted directly as with the other schedulers.  A consequence of this is the `allocation_args` map to specific call points on that python api, and thus the option pass through is not quite arbitrary.  There are 4 currently supported options for allocations which cover a majority of usecases (open an issue and let us know if one you need isn't covered!):
+As of :material-tag:`1.1.12`, the flux adapter takes advantage of new argument pass through for scheduler options that Maestro cannot abstract away.  This is done via `allocation_args` and `launcher_args` in the batch block, which expand upon the previous `args` input which only applied to `$(LAUNCHER)`.  There are some caveat's here due to the way Maestro talks to flux.  The current flux adapters all use the python api's from Flux to build the batch jobs, with the serialized batch script being serialized separately instead of submitted directly as with the other schedulers.  A consequence of this is the `allocation_args` map to specific call points on that python api, and thus the option pass through is not quite arbitrary.  There are 4 currently supported options for allocations which cover a majority of usecases (open an issue and let us know if there is a usecase you need that is not covered!):
 
 * shell options: `-o/--setopt` prefixed arguments
 * attributes: `-S/--setattr` prefixed arguments
@@ -249,7 +249,7 @@ batch:
       resource.rediscover: "true"  # Use string "true" for Flux compatibility, not "True" or bool True
   launcher_args:
     setopt:
-      optiona:   # Boolean flag, no value needed
+      optiona:   # Boolean flag, no value needed.  NOTE: This is a made up key for demonstration
 ```
 
 #### Example Batch Script
--- a/maestrowf/abstracts/interfaces/flux.py
+++ b/maestrowf/abstracts/interfaces/flux.py
@@ -248,3 +248,38 @@ class FluxInterface(ABC):
         :return: A string of the name of a FluxInterface class.
         """
         ...
+
+    @classmethod
+    def addtl_alloc_arg_types(cls):
+        """
+        Return set of additional allocation args that this adapter knows how
+        to wire up to the jobspec python apis, e.g. 'attributes',
+        'shell_options', ... This is aimed specifically at the repeated types,
+        which collect many flags/key=value pairs which go through a specific
+        jobspec call.  Everything not here gets dumped into a 'misc' group
+        for individual handling.
+
+        :return: List of string
+
+        .. note::
+
+           Should we have an enum for these or something vs random strings?
+        """
+        return []
+
+    @classmethod
+    def render_additional_args(cls, args_dict):
+        """
+        Helper to render additional argument sets to flux cli format for
+        use in constructing $(LAUNCHER) line and flux batch directives.
+        This default implementation yields a single empty string.
+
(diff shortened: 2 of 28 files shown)


docs/Maestro/scheduling.md
```diff
@@ -40,6 +44,26 @@ run locally unless at least the ``nodes`` or ``procs`` key in the step is popula
...
+* `allocation_args` for the allocation target (batch directives such as `#Flux: --setopt=foo=bar`)
+* `launcher_args` for the `$(LAUNCHER)` target (`flux run --setopt=foo=bar`)
+
+These are ~structured mappings which are designed to mimic cli usage for intuitive mapping from raw scheduler usage to Maestro.  Each of these dictionaries' keys correspond to a scheduler specific CLI argument/option or flag.  The serialization rules are as follows, with specific examples here shown for the initial implementation in the flux adapter (other schedulers will yield prefix/separator rules specific to their implementation):
+
+|  **Key Type**   | **CLI Prefix** | **Separator** | **Example YAML**    | **Example CLI Output** |
+|    :-           | :-:            |     :-:       |    :-:              |       :-               |
+|  Single letter  |  `-`           | space (`" "`) | `o: {bar: 42}`      | `-o bar=42`            |
+|  Multi-letter   | `--`           |  `=`          | `setopt: {foo: bar} | `--setopt=foo=bar`     |
+|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: } | `--setopt=foobar`      |
```
bgunnar5: Missing backticks in the 4th columns here which is making the formatting on mkdocs all wonky

maestrowf/abstracts/interfaces/schedulerscriptadapter.py
```diff
@@ -92,6 +92,36 @@ def add_batch_parameter(self, name, value):
...
+           Move this upstream into a future studystep validator?  also add
+           hooks for per scheduler normalizing of the extra args from batch
+           blocks
+        """
+        # Handle old scalar syntax which applied to allocatios only
+        if not isinstance(step_exclusive, dict):
+            return {
+                "allocation": step_exclusive,
+                "launcher": False,
+            }
```
bgunnar5: Do we want to deprecate this in the future? If so, maybe we should add a warning log statement here to notify users.

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -33,6 +34,45 @@ class FluxInterface_0490(FluxInterface):
...
+        :return: List of string
+
+        .. note::
+
+           Should we have an enum for these or something vs random strings?
+        """
+        return cls.known_alloc_arg_types
+
+    @classmethod
+    def addtl_alloc_arg_type_map(cls, option):
```
bgunnar5, line 78: May want to rename this to make it more obvious that this is a getter method. Maybe `get_addtl_alloc_arg_type`?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +86,53 @@ def get_flux_urgency(cls, urgency) -> int:
...
+                    dotstringvalue=dot_string_arg
+                )
+            # for av_name, av_value in arg_value.items():
+            #     value_str = render_arg_value(av_name, av_value)
+            #     yield "{prefix}{key}{sep}{value}".format(
+            #         prefix=arg_info['prefix'],
+            #         key=arg_key,
+            #         sep=arg_info['sep'],
+            #         value=value_str
+            #     )
```
bgunnar5: Still needed? Same question goes for `render_arg_value()` above

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -60,11 +147,19 @@ def submit(
...
         **kwargs,
     ):
-        if batch_attrs is None:
-            batch_attrs = {}
+        # Sanitize/initialize the extra batch args
+        if addtl_batch_args is None:
+            addtl_batch_args = {}
+
+        # May want to also support setattr_shell_option at some point?
+        for batch_arg_type in ["attributes", "shell_options", "conf"]:
```
bgunnar5: use `cls.addtl_alloc_arg_types` instead of hardcoded list unless you have a specific reason for this?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -79,22 +174,36 @@ def submit(
...
                     cores_per_slot=cores_per_task,
                     num_slots=procs,
                     gpus_per_slot=ngpus_per_slot,
+                    conf=conf_dict,
+                    exclusive=exclusive,
                 )
             else:
+                if conf_dict:
+                    LOGGER.warn("'conf' options not currently supported with "
+                                " nested=False.  Ignoring.")
```
bgunnar5, line 339: is this a flux limitation or a limitation we're enforcing ourselves?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -118,10 +228,26 @@ def submit(
...
+            # (-S/--setattr)
+            for batch_attr_name, batch_attr_value in addtl_batch_args["attributes"].items():
                 if batch_attr_value:
                     jobspec.setattr(batch_attr_name, batch_attr_value)
-
+                else:
+                    LOGGER.warn("Flux adapter v0.49 received null value of '%s'"
+                                " for attribute '%s'. Omitting from jobspec.",
+                                batch_attr_value,
+                                batch_attr_name)
```
bgunnar5: Being picky here but is f-string syntax not preferred nowadays?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -192,15 +319,38 @@ def parallelize(cls, procs, nodes=None, **kwargs):
...
         addtl = []
+        LOGGER.info("Processing 'exclusive': %s", exclusive)
+        if exclusive:
+            addtl.append("--exclusive")
+
         addtl_args = kwargs.get("addtl_args", {})
-        for key, value in addtl_args.items():
-            addtl.append(f"{key}={value}")
+        if addtl_args and launcher_args:
+            # TODO: better way to mark things deprecated that's not buried?
```
bgunnar5, line 463: I know Python 3.13 has a builtin `@deprecated` decorator and prior to 3.13 there's a "deprecated" pypi package that provides the same functionality. Haven't had a chance to use it yet to see if it's worth it though

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -165,6 +239,74 @@ def _convert_walltime_to_seconds(self, walltime):
...
+            arg_info = arg_info_type(arg_key)
+
+            for av_name, av_value in arg_value.items():
+                value_str = render_arg_value(av_name, av_value)
+                yield "{prefix}{key}{sep}{value}".format(
+                    prefix=arg_info['prefix'],
+                    key=arg_key,
+                    sep=arg_info['sep'],
+                    value=value_str
+                )
```
bgunnar5: same method exists in the 0_49_0 interface. Should drop this to one location unless you have a reason for this.

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,11 +95,42 @@ def __init__(self, **kwargs):
...
 
+        
+        # Setup prefixes associated with each kind of option for routing to
+        # appropriate jobspec api's
+        # Additional args can pass through on launcher only, using rule of
+        # verbose has '--' prefix, '=' flag/value separator and
+        # brief has '-' prefix, ' ' flag/value seaprator: e.g.
+        # --setattr=foo=bar, or -S foo=bar
+        self._brief_arg_info = {"prefix": "-", "sep": " "}
+        self._verbose_arg_info = {"prefix": "--", "sep": "="}
```
bgunnar5: This is also defined in the nested function of `render_additional_args`. Do we need this?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,11 +95,42 @@ def __init__(self, **kwargs):
...
+
+        # Setup known arg types
+        self._known_alloc_arg_types = ["attributes", "shell_options", "conf"]
+        self._allocation_args_map = {
+            "setopt": "shell_options",
+            "o": "shell_options",
+            "setattr": "attributes",
+            "S": "attributes",
+            "conf": "conf",
+        }
```
bgunnar5: This is also defined in 0_49_0 interface class. Do we want it here too?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,11 +95,42 @@ def __init__(self, **kwargs):
...
+
+        # setup template string that all flux cli args/batch directives use for rendering
+        self._flux_arg_str = "{prefix}{key}{sep}{value}"
+        
+        self._attr_prefixes = ['S', 'setattr']
+        self._opts_prefixes = ['o', 'setopt']
+        self._conf_prefixes = ['conf']  # No abbreviated form for this in flux docs
```
bgunnar5: are these used? Seems like the same thing as the `_allocation_args_map`

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -113,18 +144,53 @@ def __init__(self, **kwargs):
...
-            "nodes": "#INFO (nodes) {nodes}",
-            "walltime": "#INFO (walltime) {walltime}",
-            "version": "#INFO (flux adapter version) {version}",
-            "flux_version": "#INFO (flux version) {flux_version}",
-            "queue": "#INFO (queue) {queue}",
+            "nodes": f"{self._flux_directive}" + "-N {nodes}",
+            # NOTE: always use seconds to guard against upstream default behavior changes
+            "walltime": f"{self._flux_directive}" + "-t {walltime}s",
+            "queue": f"{self._flux_directive}" + "-q {queue}",
+            "bank": f"{self._flux_directive}" + "--bank {bank}",
```
bgunnar5: don't think you need '+' in these strings; can just use `f"{self._flux_directive} -N {nodes}"` and similar entries for walltime, queue, and bank

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -165,6 +239,74 @@ def _convert_walltime_to_seconds(self, walltime):
...
+            for arg_type in self._interface.addtl_alloc_arg_types()
+        }
+        for arg_key, arg_values in self._allocation_args.items():
+            # TODO: move this into a validation function for pre launch
+            #       batch args validation
+            arg_type = self._interface.addtl_alloc_arg_type_map(arg_key)
+            if arg_type is None:
+                # NO WARNINGS HERE: args in 'misc' type handled elsewhere
+                # TODO: add better mechanism for tracking whicn args
+                #       actually get used; dicts can't do this..
```
bgunnar5: dataclass use-case? 👀

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -272,14 +450,25 @@ def submit(self, step, path, cwd, job_map=None, env=None):
             LOGGER.error(msg)
             raise ValueError(msg)
 
+        # Handle the exclusive flags, updating batch block settings (default)
+        # with what's set in the step
+        step_exclusive_given = "exclusive" in step.run
+        step_exclusive = self._exclusive
+        if step_exclusive_given:
+            # Override the default with this step's setting
+            step_exclusive.update(self.get_exclusive(step.run.get("exclusive", False)))
```
bgunnar5: nearly identical code is in `get_header` and `get_parallelize_command`. Consider moving logic to a method to avoid duplication.

tests/conftest.py
```diff
@@ -143,3 +157,247 @@ def load_status_csv(file_name):
...
+
+        # Load the Specification
+        spec = YAMLSpecification.load_specification(spec_path)
+
+        environment = spec.get_study_environment()
+        steps = spec.get_study_steps()
+
+        # Set up the output directory.
+        out_dir = environment.remove("OUTPUT_PATH")
+        if output_path:
```
bgunnar5, line 211: should `output_path` be a default argument `output_path=None` since we have this check?

bgunnar5 (commented):
Not sure if this branch is totally done yet but I had some time to do a review now.

maestrowf/abstracts/interfaces/schedulerscriptadapter.py
jwhite242: That is a good question.  Should we deprecate this shorthand?  It is more often useful at the allocation level (scheduling to core scheduled partitions being one of the few use cases I can think of), so maybe useful to keep it.  But I've no strong opinions on it either way.

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: Not so much a limitation as part of the design.  There's no argument for passing it in like with the from_nest and from_batch command methods, likely because unlike those two, this one doesn't create a nested broker to attach it to.

Longer term I think this option in Maestro will may get deprecated unless we find a good use case for it; quite different behavior from the from_nest_command option.

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: not for logging: f-strings get evaluated eagerly, vs %s being lazily evaluated.  I.e. the f-string will always get processed even if this warning level gets excluded from the log handlers.    Bit of a sore spot in the otherwise excellent stdlib logging..

maestrowf/interfaces/script/fluxscriptadapter.py
jwhite242: It's more to defer them -> walltime, nodes, .. (and later queue/bank) are step specific values vs the flux_directive being adapter specific.  The + joined strings get rendered later when writing out the batch script for individual steps.

docs/Maestro/scheduling.md
```diff
@@ -40,6 +44,28 @@ run locally unless at least the ``nodes`` or ``procs`` key in the step is popula
...
+* `launcher_args` for the `$(LAUNCHER)` target (`flux run --setopt=foo=bar`)
+
+These are ~structured mappings which are designed to mimic cli and batch script directive syntax for intuitive mapping from raw scheduler usage to Maestro.  Each of these dictionaries' keys correspond to a scheduler specific CLI argument/option or flag.  The serialization rules are as follows, with specific examples here shown for the initial implementation in the flux adapter (other schedulers will yield prefix/separator rules specific to their implementation):
+
+|  **Key Type**         | **Prefix** | **Separator** | **Example YAML**     | **Example CLI Input/Directive** |
+|    :-                 | :-            | :-       |    :-                |       :-               |
+|  Single letter        |  `-`           | `" "` (space) | `o: {bar: 42}`       | `-o bar=42`            |
+|  Multi-letter         | `--`           |  `=`          | `setopt: {foo: bar}` | `--setopt=foo=bar`     |
+|  Boolean flag w/key   | as above       | as above      | `setopt: {foobar: }` | `--setopt=foobar`      |
+|  Boolean flag w/o key | as above       | as above      | `exclusive: `        | `--exclusive`      |
```
bgunnar5: In the `Example YAML` Column here, you may want each of these to be multi-line instead like users are used to seeing? I've no strong opinion on this, just thinking from users POV.

I've done this in a table before with html tags. So for that first example it'd become:

```
<pre><code><span>o:</span></br><span>  bar:</span></br><span>    42</span></code></pre>
```

Ugly in raw text but formats nicely in the table.

docs/Maestro/scheduling.md
```diff
@@ -187,6 +213,64 @@ See the [flux framework](https://flux-framework.readthedocs.io/en/latest/index.h
     The Flux scheduler itself and Maestro's flux adapter are still in a state of flux and may go through breaking changes more frequently than the Slurm and LSF scheduler adapters.
    
 
+### Extra Flux Args
+----
+
+As of :material-tag:`1.1.12`, the flux adapter takes advantage of new argument pass through for scheduler options that Maestro cannot abstract away.  This is done via `allocation_args` and `launcher_args` in the batch block, which expand upon the previous `args` input which only applied to `$(LAUNCHER)`.  There are some caveat's here due to the way Maestro talks to flux.  The current flux adapters all use the python api's from Flux to build the batch jobs, with the serialized batch script being serialized separately instead of submitted directly as with the other schedulers.  A consequence of this is the `allocation_args` map to specific call points on that python api, and thus the option pass through is not quite arbitrary.  There are 4 currently supported options for allocations which cover a majority of usecases (open an issue and let us know if one you need isn't covered!):
```
bgunnar5:
```suggestion
As of :material-tag:`1.1.12`, the flux adapter takes advantage of new argument pass through for scheduler options that Maestro cannot abstract away.  This is done via `allocation_args` and `launcher_args` in the batch block, which expand upon the previous `args` input which only applied to `$(LAUNCHER)`.  There are some caveat's here due to the way Maestro talks to flux.  The current flux adapters all use the python api's from Flux to build the batch jobs, with the serialized batch script being serialized separately instead of submitted directly as with the other schedulers.  A consequence of this is the `allocation_args` map to specific call points on that python api, and thus the option pass through is not quite arbitrary.  There are 4 currently supported options for allocations which cover a majority of usecases (open an issue and let us know if there's one you need that isn't covered!):
```

docs/Maestro/scheduling.md
```diff
@@ -187,6 +213,64 @@ See the [flux framework](https://flux-framework.readthedocs.io/en/latest/index.h
...
+      foo: bar
+    o:
+      bar: 42
+    setattr:
+      foobar: "whoops"
+    conf:
+      resource.rediscover: "true"  # Use string "true" for Flux compatibility, not "True" or bool True
+  launcher_args:
+    setopt:
+      optiona:   # Boolean flag, no value needed
```
bgunnar5: is this supposed to be "option a" or "optional"?

docs/Maestro/scheduling.md
```diff
@@ -187,6 +213,64 @@ See the [flux framework](https://flux-framework.readthedocs.io/en/latest/index.h
...
+``` console
+#flux: -q debug
+#flux: --bank=guests
+#flux: -t 300s
+#flux: --setopt=foo=bar
+#flux: --setopt=bar=42
+#flux: --setattr=foobar=whoops
+#flux: --conf=resource.rediscover=true
+
+flux run -n 1 -N 1 -c 1 --setopt=optiona  myapplication
```
bgunnar5, line 268: same question, "option a" or "optional"? From this context I'm assuming the former in which case we may want to change the naming convention for clarity?

maestrowf/abstracts/interfaces/flux.py
```diff
@@ -248,3 +248,88 @@ def key(self):
...
+        'shell_options', ... This is aimed specifically at the repeated types,
+        which collect many flags/key=value pairs which go through a specific
+        jobspec call.  Everything not here gets dumped into a 'misc' group
+        for individual handling.
+
+        :return: List of string
+
+        .. note::
+
+           Should we have an enum for these or something vs random strings?
```
bgunnar5, line 266: I like this idea. Could probably help with safeguarding against unsupported options

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -33,6 +41,74 @@ class FluxInterface_0490(FluxInterface):
...
+           by default in flux, attributes.system?)
+        """
+        if arg_type in cls.known_alloc_arg_types:
+            return cls._addtl_arg_cli_key.get(arg_type)
+
+        elif arg_type in cls._addtl_alloc_arg_type_map:
+            return cls._addtl_arg_cli_key.get(cls._addtl_alloc_arg_type_map[arg_type])
+
+        else:
+            return arg_type
```
bgunnar5, line 113: would this be cleaner to just say:

```python
# See if we're using an alias for a known argument
alloc_arg_type = cls._addtl_alloc_arg_type_map.get(arg_type, arg_type)

# If the argument is known, retrieve the appropriate cli key. Otherwise, return the key itself
return cls._addtl_arg_cli_key.get(alloc_arg_type, alloc_arg_type)
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +122,125 @@ def get_flux_urgency(cls, urgency) -> int:
...
+        base_render_tmpl = "{prefix}{cli_key}{sep}{dotpath}"
+        value_render_tmpl = "={value}"
+        for arg_key, arg_value in args_dict.items():
+            # Get the cli key and associated rendering info
+            cli_key = cls.get_addtl_arg_cli_key(arg_key)
+            cli_info = cls.get_cli_arg_prefix_sep(cli_key)
+
+            # Note: dotpath encoding comes after the group/prefix (setattr, ...)
+            for dotpath, value in iter_dotpath_items(arg_value):
+                # print(f"{dotpath=} {value=}")
```
bgunnar5:
```suggestion
                LOGGER.debug(f"{dotpath=} {value=}")
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +122,125 @@ def get_flux_urgency(cls, urgency) -> int:
...
+                rendered_opt = base_render_tmpl.format(
+                    prefix=cli_info['prefix'],
+                        cli_key=cli_key,
+                        sep=cli_info['sep'],
+                        dotpath=dotpath
```
bgunnar5:
```suggestion
                    cli_key=cli_key,
                    sep=cli_info['sep'],
                    dotpath=dotpath
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +122,125 @@ def get_flux_urgency(cls, urgency) -> int:
...
+        Helper to normalize additional arguments to known types and an
+        unflattened nested dictionary structure.  This unflattens any
+        dotpath encoded nested dictionary keys.
+
+        :param args_dict: Dictionary of flux arg keys and name: value pairs
+        :return: dict of packed args with top level keys being the adapter
+                 version specific addtl_alloc_arg_types
+        """
+        # First, normalize and unflatten everything into dicts
+        # print(f"original_additional_args: {args_dict=}")
```
bgunnar5:
```suggestion
        LOGGER.debug(f"original_additional_args: {args_dict=}")
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +122,125 @@ def get_flux_urgency(cls, urgency) -> int:
...
+                # unflattened_batch_args[arg_key] = arg_values
+
+            # Update to ensure we don't clobber prior values
+            # print(f"pre-update  unflattened_batch_args: {unflattened_batch_args=}")
+            unflattened_batch_args = update_recursive(unflattened_batch_args,
+                                                      unflattened_batch_arg)
+        #     print(f"post-update unflattened_batch_args: {unflattened_batch_args=}")
+
+        # print(f"normalized_additional_args: {unflattened_batch_args=}")
+        return unflattened_batch_args
```
bgunnar5, line 228: Changing commented print statements to logger.debug statements and removing commented code that's not necessary.
```suggestion
        for arg_key, arg_values in args_dict.items():
            arg_type = cls.addtl_alloc_arg_type_map(arg_key)
            LOGGER.debug(f"normalize_additional_args: {arg_key=} -> {arg_type=}")
            if arg_type is None:
                # We only restrict for allocations, so defer logging/rejection
                # of types not handled by jobspec builder
                arg_type = arg_key

            if isinstance(arg_values, dict):
                unflattened_batch_arg = {arg_type: unflatten_dotpath_dict(arg_values)}
            else:
                unflattened_batch_arg = {arg_type: arg_values}

            # Update to ensure we don't clobber prior values
            LOGGER.debug(f"pre-update  unflattened_batch_args: {unflattened_batch_args=}")
            unflattened_batch_args = update_recursive(unflattened_batch_args,
                                                      unflattened_batch_arg)
            LOGGER.debug(f"post-update unflattened_batch_args: {unflattened_batch_args=}")

        LOGGER.debug(f"normalized_additional_args: {unflattened_batch_args=}")
        return unflattened_batch_args
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -46,6 +122,125 @@ def get_flux_urgency(cls, urgency) -> int:
...
+        packed_batch_args = {
+            arg_type: {}
+            for arg_type in cls.addtl_alloc_arg_types()
+        }
+
+        dotpath_format = ["attributes", "shell_options"]
+
+        for arg_key, arg_values in args_dict.items():
+            coerced_vals = coerce_dict_values(arg_values,
+                                              lambda x: 1 if x is None else x)
```
bgunnar5, line 258: kinda unbelievable that this is required and flux python doesn't handle this for us

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -79,22 +284,39 @@ def submit(
             # for a single node, don't use a broker -- but introduce a flag
             # that can force a single node to run in a broker.
 
+            # Attach any conf inputs to the jobspec
+            conf_dict = addtl_batch_args.get('conf', None)
+            # if "conf" in addtl_batch_args:
+            #     conf_dict = addtl_batch_args['conf']
+
+            # else:
+            #     conf_dict = None
```
bgunnar5:
```suggestion
            # Attach any conf inputs to the jobspec
            conf_dict = addtl_batch_args.get('conf', None)
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
bgunnar5: got it, that makes sense

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -117,11 +342,34 @@ def submit(
...
+            # Add in job shell options if not null (-o/--setopt)
+            # NOTE: these are sanitized upstream to be dotpath, value tuples
+            for batch_opt_dotpath, batch_opt_value in addtl_batch_args["shell_options"]:
+                # if batch_opt_value:
+                jobspec.setattr_shell_option(batch_opt_dotpath, batch_opt_value)
+                # else:
+                #     LOGGER.warn("Flux adapter v0.49 received null value of '%s'"
+                #                 " for shell option '%s'. Omitting from jobspec.",
+                #                 batch_opt_value,
+                #                 batch_opt_dotpath)
```
bgunnar5:
```suggestion
            for batch_attr_dotpath, batch_attr_value in addtl_batch_args["attributes"]:
                jobspec.setattr(batch_attr_dotpath, batch_attr_value)

            # Add in job shell options if not null (-o/--setopt)
            # NOTE: these are sanitized upstream to be dotpath, value tuples
            for batch_opt_dotpath, batch_opt_value in addtl_batch_args["shell_options"]:
                jobspec.setattr_shell_option(batch_opt_dotpath, batch_opt_value)
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -192,15 +441,38 @@ def parallelize(cls, procs, nodes=None, **kwargs):
...
+        if 'o' in launcher_args:
+            launcher_args['o'].update(**addtl_args)
+        else:
+            launcher_args['o'] = addtl_args
+
+        addtl += [arg for arg in cls.render_additional_args(launcher_args)]
+        args.extend(addtl)
+        # for key, value in addtl_args.items():
+
+        #     addtl.append(f"{key}={value}")
```
bgunnar5:
```suggestion
```

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -192,15 +441,38 @@ def parallelize(cls, procs, nodes=None, **kwargs):
...
 
-        if addtl:
-            args.append("-o")
-            args.append(",".join(addtl))
+        # if addtl:
+        #     args.append("-o")
+        #     args.append(",".join(addtl))
```
bgunnar5:
```suggestion
```

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,11 +97,26 @@ def __init__(self, **kwargs):
...
+        # Normalize the allocation and launcher args to a flat dict, removing dotpath encoding
+        # TODO: workon a validation api that can be hooked up to yamlspec ingestion machinery for
+        # early error messaging
+        if self._allocation_args:
+            self._allocation_args = self._interface.normalize_additional_args(self._allocation_args)
+            # pprint(f"{self._allocation_args=}")
+
+        if self._launcher_args:
+            self._launcher_args = self._interface.normalize_additional_args(self._launcher_args)
+            # pprint(f"{self._launcher_args=}")
```
bgunnar5:
```suggestion
        if self._allocation_args:
            self._allocation_args = self._interface.normalize_additional_args(self._allocation_args)

        if self._launcher_args:
            self._launcher_args = self._interface.normalize_additional_args(self._launcher_args)
```

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,11 +97,26 @@ def __init__(self, **kwargs):
         # NOTE: Host doesn"t seem to matter for FLUX. sbatch assumes that the
```
bgunnar5, line 96: This `__init__` method is getting pretty large. May want to consider splitting into separate methods in the future

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -111,20 +128,56 @@ def __init__(self, **kwargs):
...
+        # Check for the flag in additional args and pop it off, letting step key win later
+        # NOTE: only need presence of key as this mimics cli like flag behavior
+        # TODO: promote this to formally supported behavior for all adapters
+        exclusive_keys = ['x', 'exclusive']
+        if all(ekey in self._allocation_args for ekey in exclusive_keys):
+            LOGGER.warn("Repeated addition of exclusive flags 'x' and 'exclusive' in allocation_args.")
+
+        alloc_eflags = [self._allocation_args.pop(ekey, None) for ekey in exclusive_keys]
+        if alloc_eflags:
+            if step_exclusive:
```
bgunnar5: should this be `if step_exclusive and self._exclusive['allocation']` like the launcher section below? If so, you may just want a common method for this shared logic

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -272,14 +413,54 @@ def submit(self, step, path, cwd, job_map=None, env=None):
...
+
+        # if self._batch["queue"]:
+        #     normalized_alloc_args["attributes"]["system"]["queue"] = self._batch["queue"]
+        # if self._batch["bank"]:
+        #     # TODO: revisit whether it makes sense to add bank if queue is empty ->
+        #     #       nested brokers usually have neither, and bank falls through silently..
+        #     normalized_alloc_args["attributes"]["system"]["bank"] = self._batch["bank"]
+
+        # pprint(f"Packed alloc args for {step.name}:")
+        # pprint(packed_alloc_args)
```
bgunnar5:
```suggestion
        # Normalize the allocation args to api flux.job.JobspecV1 expects
        packed_alloc_args = self._interface.pack_addtl_batch_args(self._allocation_args)

        # Add queue and bank
        queue = self._batch["queue"]
        if queue == "":
            queue = None
        bank = self._batch["bank"]
        if bank == "":
            bank = None

```

bgunnar5 (approved):
Looks good. A lot of these suggestions are just removing commented code and print statements. Really nothing major

docs/Maestro/scheduling.md
jwhite242: optiona; just a made up name for some random option to demonstrate the syntax mapping from yaml to batch script

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: oof..   sorry about missing all this stray refactoring/debugging detritus, and thanks for catching it!

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: It does, just not in the python api.  Their cli machinery upstream of this python api does most of the format conversion that this pr ended up having to add back on (it's where i found this normalization).  I'd settle for better documentation on what these args should be in the api instead of having to trace all the call points to find where and if the conversions happen

maestrowf/interfaces/script/fluxscriptadapter.py
jwhite242: thanks for commenting on this one..  looking through the rest of the exclusive handling it appears these new modes aren't actually tested either which is troublesome, so let me go add that and verify the right thing's happening here..  the behaviors are a little different given old method exclusive = allocation only..    Might defer some of the refactoring to one of the followon's.  I think more of this needs to be in the script adapter base class and just do more normalization upstream, and get it setup for a more layered approach to add on the step keys in the header/writescript/submit funcs and only handle the batch block source in the init..    ugh, i've made a mess, lol..

bgunnar5 (approved):
Great work!

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: yeah, that's probably good for this version.  In the subsequent PR's that will generalize this, something a little more descriptive like 'get_normalized..'  or something?  though with some iteration to make it not 30 char's long..

jwhite242:
> Great work!

Thanks forcing a look at that disaster around the exclusive behavior changes!

PR #460, opened 2025-12-04
Add controls for attaching nodes/task info to jobspecs and launchers

jwhite242:
Adds additional control on how resource specs are attached to both batch jobs and launchers separately.  Enables use of more dynamic resource configuration such as the gpumode introduced with flux for modern AMD machines which can change the number of logical gpus after job scheduling time.  Pre 1.1.12 behavior attaches tasks to the jobspec in flux, which can result in 'unsatisfiable' job errors due to there not being sufficient logical gpu's at jobspec validation time to fulfill the job request.  This case requires not binding the tasks to the jobspec, but still binding the tasks to the $(LAUNCHER) generated flux run.  

This initial version adds support for flux only; slurm/lsf/etc handling to follow in subsequent release (will be no-op everywhere but flux scheduled jobs).


jwhite242:
Closing this for now in favor of #467 to default to single slot per node configurations.  Can revisit this idea later if more granular control of slots on the node/allocation level is needed.

PR #465, opened 2026-01-27
Add compact, sortable 'hashing' strategy for step ids/workspaces

jwhite242:
Implements MEP 003's proposal for more human readable hashing of workspaces/step id's.  Replaces default 'md5' algorithm for hashing step combinations in favor of sortable/human readable 'hash'.  This is not a true 'hash', in that it needs all combos for a step to compute the pattern.  Pattern is: `step_name_instance_<number>`, where <number> is a zero padded integer (zero based index) based on the number of used step combinations for a particular step.  Zero padding ensures proper sorting on filesystems to improve usability.

* replaces original md5 hash algorithm
* reuses `--hashws` cli option to activate
* adds local integration test
* adds test fixtures for checking workspace layouts relative to study root (can't hardcode absolute paths in tests)
* adds test fixtures for checking substituted workspace tokens in generated step scripts.


tests/integration/test_local.py
```diff
@@ -0,0 +1,336 @@
...
+from maestrowf.specification.yamlspecification import YAMLSpecification
+from rich.console import Console
+console = Console()
+
+# Tag every test in this file as requiring flux
+pytestmark = [pytest.mark.sched_local,
+              pytest.mark.integration,]
+
+
+# @pytest.mark.parametrize(
```
doutriaux1: Is this commented out on purpose?

tests/integration/test_local.py
```diff
@@ -0,0 +1,336 @@
...
+    console.rule("stdout")
+    console.print(spec_results.stdout)
+    console.rule("stderr")
+    console.print(spec_results.stderr)
+
+    """
+    Demonstrates how to display a tree of files / directories with the Tree renderable.
+    """
+
+    # import os
```
doutriaux1: clean this up?

tests/integration/test_local.py
```diff
@@ -0,0 +1,336 @@
...
+                        path_pattern=rf'{re.escape(line_prefix)}\s+(?P<path>[^"\s]+)"?'
+                    )
+                    console.print(f"{rel_path=}, type: {type(rel_path)}.  {expected_rel_path=}, type: {type(expected_rel_path)}")
+                    assert rel_path == expected_rel_path, (
+                        f"In {gen_script_file}, for prefix {line_prefix!r}, "
+                        f"expected relative path {expected_rel_path}, got {rel_path}"
+                    )
+
+    # Dummy test to force failure and show outputs
+    # assert completed_successfully == 'NOT A REAL STATUS'
```
doutriaux1, line 248: debug?

doutriaux1 (approved):
Looks good, only suggestion I have that can be ignored or added in a later PR would be to allow to control the hashing from within the yaml file either via key or env variable.

jwhite242:
> Looks good, only suggestion I have that can be ignored or added in a later PR would be to allow to control the hashing from within the yaml file either via key or env variable.

Yeah, that's gonna be a whole other pr.  There's currently no where to put any of the study config in spec and I need to explore some alternatives to cluttering up the env/variables block with it as things like that + autoyes/rlimit/sleep have a different purpose from variables/labels/etc

tests/integration/test_local.py
jwhite242: have to do that if you want to see any of the printout's: pytest eats everything unless there's a failure.  Really need some alternative switch in here that doesn't leave this kind of debris behind for a nicer test building/debugging workflow...    But not on this pr, so will delete all these debug outputs

PR #467, opened 2026-03-12
Update flux exclusive behavior modify task info passed to allocation/launcher

jwhite242:
Updates handling of exclusive keys to modify the batch directives and launcher rendering

* Prunes everything but nodes from allocations
* If set on launcher, overrides any cores per task, gpu flags, leaving nodes and tasks only
* Jobspec creation via Python api now sets num_slots = nodes when exclusive is set, matching flux behavior

Updates default renderings:
* cores per task omitted from launcher if not set in step keys
* removes default 'nodes=1' in batch header/directives to sync with step keys
* retains required >=1 for tasks/cores per task in jobspec creation but omits from rendered batch scripts to sync with flux default behaviors

Unblocks scheduling to nodes with configurable resources: subdividing gpus happens after allocation startup, and prior behavior of attaching tasks/gpus to allocation would cause flux to reject such jobs (unsatisfiable).

### DIFF
--- a/maestrowf/interfaces/script/_flux/flux0_49_0.py
+++ b/maestrowf/interfaces/script/_flux/flux0_49_0.py
@@ -423,7 +423,7 @@ class FluxInterface_0490(FluxInterface):
     @classmethod
     def parallelize(cls, procs, nodes=None, launcher_args=None, **kwargs):
 
-        args = ["flux", "run"]  #, "-n", str(procs)]
+        args = ["flux", "run"]
         
         exclusive = kwargs.pop('exclusive', False)
         
@@ -431,9 +431,6 @@ class FluxInterface_0490(FluxInterface):
         if nodes is not None and nodes != '':
             args.append("-N")
             args.append(str(nodes))
-        # ntasks = nodes if nodes else 1
-        # args.append("-N")
-        # args.append(str(ntasks))
 
         # NOTE: should we raise an exception here instead of just logging the error?
         #       better pre-run validation might be more useful...
@@ -452,17 +449,11 @@ class FluxInterface_0490(FluxInterface):
         
         if cores_per_task is not None and cores_per_task != '' and not exclusive:
             args.append("-c")
-            # Error checking -> more comprehensive handling in base schedulerscriptadapter?
-            # if not kwargs["cores per task"]:
-            #     cores_per_task = 1
-            # else:
-            #     cores_per_task = kwargs["cores per task"]
-
-            # args.append(str(kwargs["cores per task"]))
+            # TODO: more comprehensive handling in base schedulerscriptadapter?
             args.append(str(cores_per_task))
 
             LOGGER.info("Adding 'cores per task' %s to flux args",
-                        str(kwargs["cores per task"]))
+                        str(cores_per_task))
 
         ngpus = kwargs.get("gpus", 0)
         if ngpus and ngpus != '' and not exclusive:
--- a/maestrowf/interfaces/script/_flux/flux0_49_0.py
+++ b/maestrowf/interfaces/script/_flux/flux0_49_0.py
@@ -321,7 +321,12 @@ class FluxInterface_0490(FluxInterface):
                 )
                 # Need to attach broker opts to the constructor?
                 # TODO: Add in extra broker options if not null
-                ngpus_per_slot = int(ceil(ngpus / nodes))
+                # ngpus_per_slot = int(ceil(ngpus / nodes))
+                if nodes:
+                    ngpus_per_slot = int(ceil(ngpus / nodes))
+                else:
+                    ngpus_per_slot = None
+                LOGGER.warn(f"NODES recieved: {nodes}")
                 jobspec = flux.job.JobspecV1.from_nest_command(
                     [path],
                     num_nodes=nodes,
@@ -418,20 +423,40 @@ class FluxInterface_0490(FluxInterface):
     @classmethod
     def parallelize(cls, procs, nodes=None, launcher_args=None, **kwargs):
 
-        args = ["flux", "run", "-n", str(procs)]
-
+        args = ["flux", "run"]  #, "-n", str(procs)]
+        
+        exclusive = kwargs.pop('exclusive', False)
+        
         # if we've specified nodes, add that to wreckrun
-        ntasks = nodes if nodes else 1
-        args.append("-N")
-        args.append(str(ntasks))
+        if nodes is not None and nodes != '':
+            args.append("-N")
+            args.append(str(nodes))
+        # ntasks = nodes if nodes else 1
+        # args.append("-N")
+        # args.append(str(ntasks))
 
-        if "cores per task" in kwargs:
+        # NOTE: should we raise an exception here instead of just logging the error?
+        #       better pre-run validation might be more useful...
+        #       or should we just prune the exclusive flag and run it anyway?
+        if procs and (nodes is None or nodes == '') and exclusive:
+            LOGGER.error(
+                "Invalid use of exclusive on launcher: "
+                "exclusive can only be set with a node count"
+            )
+
+        if procs:
+            args.append("-n")
+            args.append(str(procs))
+
+        cores_per_task = kwargs.get("cores per task", None)
+        
+        if cores_per_task is not None and cores_per_task != '' and not exclusive:
             args.append("-c")
             # Error checking -> more comprehensive handling in base schedulerscriptadapter?
-            if not kwargs["cores per task"]:
-                cores_per_task = 1
-            else:
-                cores_per_task = kwargs["cores per task"]
+            # if not kwargs["cores per task"]:
+            #     cores_per_task = 1
+            # else:
+            #     cores_per_task = kwargs["cores per task"]
 
             # args.append(str(kwargs["cores per task"]))
             args.append(str(cores_per_task))
@@ -440,7 +465,7 @@ class FluxInterface_0490(FluxInterface):
                         str(kwargs["cores per task"]))
 
         ngpus = kwargs.get("gpus", 0)
-        if ngpus:
+        if ngpus and ngpus != '' and not exclusive:
             gpus = str(ngpus)
             args.append("-g")
             args.append(gpus)
@@ -451,8 +476,6 @@ class FluxInterface_0490(FluxInterface):
             launcher_args = {}
 
         # Look for optional exclusive flag
-        exclusive = kwargs.pop('exclusive', False)
-
         addtl = []
         LOGGER.info("Processing 'exclusive': %s", exclusive)
         if exclusive:
@@ -461,7 +484,7 @@ class FluxInterface_0490(FluxInterface):
         addtl_args = kwargs.get("addtl_args", {})
         if addtl_args and launcher_args:
             # TODO: better way to mark things deprecated that's not buried?
-            LOGGER.warn("'args' input is deprecated in v1.1.12.  Use the more "
+            LOGGER.warn("'args' input is deprecated in v1.2.  Use the more "
                         "flexible 'launcher_args' going forward. Combining.")
         if 'o' in launcher_args:
             launcher_args['o'].update(**addtl_args)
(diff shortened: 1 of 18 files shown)


maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -321,7 +321,12 @@ def submit(
                 )
                 # Need to attach broker opts to the constructor?
                 # TODO: Add in extra broker options if not null
-                ngpus_per_slot = int(ceil(ngpus / nodes))
+                # ngpus_per_slot = int(ceil(ngpus / nodes))
+                if nodes:
+                    ngpus_per_slot = int(ceil(ngpus / nodes))
+                else:
+                    ngpus_per_slot = None
+                LOGGER.warn(f"NODES recieved: {nodes}")
```
doutriaux1: typo and also why is it a wraning? Shouldn't this be an info? (unless you want it in the else)

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -418,20 +423,40 @@ def submit(
     @classmethod
     def parallelize(cls, procs, nodes=None, launcher_args=None, **kwargs):
 
-        args = ["flux", "run", "-n", str(procs)]
-
+        args = ["flux", "run"]  #, "-n", str(procs)]
```
doutriaux1: do we want to keep the commented out bit?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -418,20 +423,40 @@ def submit(
...
+        exclusive = kwargs.pop('exclusive', False)
+        
         # if we've specified nodes, add that to wreckrun
-        ntasks = nodes if nodes else 1
-        args.append("-N")
-        args.append(str(ntasks))
+        if nodes is not None and nodes != '':
+            args.append("-N")
+            args.append(str(nodes))
+        # ntasks = nodes if nodes else 1
```
doutriaux1: same comment do we keep this?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -418,20 +423,40 @@ def submit(
...
+        # args.append("-N")
+        # args.append(str(ntasks))
+
+        # NOTE: should we raise an exception here instead of just logging the error?
+        #       better pre-run validation might be more useful...
```
doutriaux1, line 436: Yes I agree I'd raise an error since we log it, or downgrade the log to a warning.

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -418,20 +423,40 @@ def submit(
...
+        #       or should we just prune the exclusive flag and run it anyway?
+        if procs and (nodes is None or nodes == '') and exclusive:
+            LOGGER.error(
+                "Invalid use of exclusive on launcher: "
+                "exclusive can only be set with a node count"
+            )
+
+        if procs:
+            args.append("-n")
+            args.append(str(procs))
```
doutriaux1, line 446: `args.append(f"- n {procs}")` ?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -418,20 +423,40 @@ def submit(
...
 
-        if "cores per task" in kwargs:
+        cores_per_task = kwargs.get("cores per task", None)
```
doutriaux1, line 448: this is where we could add the magical function that understands `core per task` `Cores_per-tasks` etc... 😄

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -440,7 +465,7 @@ def parallelize(cls, procs, nodes=None, launcher_args=None, **kwargs):
                         str(kwargs["cores per task"]))
 
         ngpus = kwargs.get("gpus", 0)
```
doutriaux1, line 458: 0 for default? not '' ?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, **kwargs):
 
         # NOTE: Host doesn"t seem to matter for FLUX. sbatch assumes that the
         # current host is where submission occurs.
-        self.add_batch_parameter("nodes", kwargs.pop("nodes", "1"))
+        # self.add_batch_parameter("nodes", kwargs.pop("nodes", "1"))
```
doutriaux1: remove commented line?

maestrowf/interfaces/script/_flux/flux0_49_0.py
```diff
@@ -461,7 +484,7 @@ def parallelize(cls, procs, nodes=None, launcher_args=None, **kwargs):
         addtl_args = kwargs.get("addtl_args", {})
         if addtl_args and launcher_args:
             # TODO: better way to mark things deprecated that's not buried?
-            LOGGER.warn("'args' input is deprecated in v1.1.12.  Use the more "
+            LOGGER.warn("'args' input is deprecated in v1.2.  Use the more "
```
doutriaux1, line 478: does `LOGGER.warn` also raise a warning? Just wondering in case someone calls this from python and treats warnings as errors.

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -264,9 +264,25 @@ def get_header(self, step):
...
             str(self._convert_walltime_to_seconds(walltime))
 
-        if run["nodes"]:
+        # Unpack exclusive to modify procs/core per task
+        step_exclusive = step.run.get("exclusive", None)
+
+        step_exclusive = self.resolve_exclusive(self._exclusive, step_exclusive)
+
+        nodes = run.get("nodes", None)
+        LOGGER.info("Retrived 'nodes' from the step: '%s'", nodes)
```
doutriaux1: typo

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -264,9 +264,25 @@ def get_header(self, step):
...
+        print(f"INFO: retrieved 'nodes' from step: '{nodes}'")
```
doutriaux1: LOGGER and print ?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -264,9 +264,25 @@ def get_header(self, step):
...
+        if nodes is not None and nodes != '':  # catch nodes = 0 here?
```
doutriaux1: probably a good idea indeed to catch nodes=0 somewhere

doutriaux1: or does node=0 as a special meaning?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -322,7 +331,8 @@ def get_parallelize_command(self, procs, nodes=None, **kwargs):
         :returns: A string of the parallelize command configured using nodes
                   and procs.
         """
-        ntasks = nodes if nodes else self._batch.get("nodes", 1)
+        # nodes = nodes
```
doutriaux1: cleanup?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -350,14 +360,18 @@ def submit(self, step, path, cwd, job_map=None, env=None):
...
-            if not nodes:
-                nodes = 1
-            else:
-                nodes = int(nodes)
+        nodes = step.run.get("nodes", None)
+        processors = step.run.get("procs", 1)
+
+        if nodes == '':
+            nodes = None
+        if nodes is not None and nodes != '':  # Have to account for empty string!
```
doutriaux1: i dont thinks `nodes==''` is possible at this point considering previous line

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -350,14 +360,18 @@ def submit(self, step, path, cwd, job_map=None, env=None):
...
+            nodes = int(nodes)
+        # if not isinstance(nodes, int):
```
doutriaux1: cleanup?

maestrowf/interfaces/script/fluxscriptadapter.py
```diff
@@ -394,7 +408,7 @@ def submit(self, step, path, cwd, job_map=None, env=None):
             raise val_error
 
         # Calculate nprocs
-        ncores = cores_per_task * nodes
+        ncores = cores_per_task * processors  # What was this check doing?
```
doutriaux1, line 412: seems like a big change, was it broken before?

tests/data/specs/hello_bye_parameterized_flux_2.yaml
```diff
@@ -16,12 +16,12 @@ batch:
         gpumode: CPX
       conf:
         resource:
-          rediscover: True # Flux wants 'true', not 'True' that python bool serializes to
+          rediscover: True
```
doutriaux1, line 19: removed the comment?

tests/data/specs/hello_bye_parameterized_flux_2.yaml
```diff
@@ -16,12 +16,12 @@ batch:
...
           noverify: True
         
     launcher_args:
       setopt:
-        fastload:
+        fastload:
```
doutriaux1, line 24: want the new space?

tests/data/specs/hello_bye_parameterized_flux_3.yaml
```diff
@@ -16,7 +16,7 @@ batch:
         gpumode: CPX
       conf:
         resource:
-          rediscover: "true" # Flux wants 'true', not 'True' that python bool serializes to
+          rediscover: "true"
```
doutriaux1, line 19: so string here but `True` above? Is it a flux issue?

doutriaux1 (approved):

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: convention on all the adapters so far is each arg as a separate list item, following some of the subprocessing style arguments for shell commands, so going to leave it this way to be consistent

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: 0 makes more sense i think: the '' is an artifact of some processing upstream that i want to fix later (all these != '' tests in here are silly)

maestrowf/interfaces/script/_flux/flux0_49_0.py
jwhite242: well, if they treat warnings as errors then this would stop them and force the update as intended ;)

sadly it doesn't look like logger triggers that behavior on it's own; warn is just a severity level in the logs

maestrowf/interfaces/script/fluxscriptadapter.py
jwhite242: just a little present for jumping the gun and reviewing it while still marked 'draft' ;)

maestrowf/interfaces/script/fluxscriptadapter.py
jwhite242: nodes=0 doesn't mean anything different than None or '';  kind of shocked users haven't tried '0' yet

maestrowf/interfaces/script/fluxscriptadapter.py
jwhite242: allowing nodes to be 0 or None (never could be before) broke it. 

will likely remove it entirely later.. not clear this check is useful given there's no place to eval against the machine...  might only catch negative values now, which will be better checked upstream after future refactoring.

tests/data/specs/hello_bye_parameterized_flux_2.yaml
jwhite242: stray debris from prior pr: already fixed it so it's not a user facing problem anymore

tests/data/specs/hello_bye_parameterized_flux_3.yaml
jwhite242: either one works: it's json syntax.  see 'render_additional_args' in the flux0_49_0.py for explanation

Issue #476, opened 2026-05-29
Maestro write/update status.csv and check for cancellation after every job step


atgrier:
Currently the `status.csv` is not written (such as for synchronous login node jobs) until all of the first job step jobs are completed, nor can cancellation be issued or checked until that point. Proposed changes:

- Write `status.csv` immediately after the study graph is created
- Update `status.csv` after synchronous login node job are completed
- Additionally check for cancellation between synchronous login node jobs

atgrier:
@jwhite242 I believe I've solved this without too many changes. Simply writing the `status.csv` immediately after the study graph is constructed is as simple as adding a `self._exec_dag.write_status()` call at the end of `Conductor.initialize()`. Updating it between synchronous login node jobs and the cancellation improvements are a little more involved but still relatively simple.

PR #477, opened 2026-05-29
write status.csv as soon as study graph is constructed and between sy…

atgrier:
This PR addresses #476 

Currently the status.csv is not written (such as for synchronous login node jobs) until all of the first job step jobs are completed, nor can cancellation be issued or checked until that point. Proposed changes:

- Write status.csv immediately after the study graph is created
- Update status.csv after synchronous login node job are completed
- Additionally check for cancellation between synchronous login node jobs


maestrowf/datastructures/core/executiongraph.py
```diff
@@ -969,6 +1039,12 @@ def execute_ready_steps(self):
             LOGGER.debug("Launching job %d -- %s", i, _record.name)
             self._execute_record(_record, adapter)
 
+            # Write status after each job execution to provide incremental
+            # updates, especially for login node jobs that complete
+            # synchronously.
+            if self._status_path:
+                self.write_status()
```
jwhite242: So I think this (and the cancellation) machinery should remain out at the conductor level, not in the job submission hot-loop inside the execution graph.  Bench-marking the submission speed fix that's going into this release has the new floor at ~20-30 job submissions per second, and hitting the file system for this and cancellation that often is likely to cause problems (might even make the status cmd fail outright while jobs are being submitted due to io contention).

I've got an alternate local adapter + executor in the works to make it behave more like flux/slurm/etc schedulers (i.e. non-blocking) that should be a better path making things more responsive.

atgrier: I reverted the bulk of the changes, including the test file that was mostly for the cancellation changes. Now it's limited to a single addition of a status write at conclusion of initializing the conductor, along with a couple minor clerical edits.

maestrowf/conductor.py
```diff
@@ -361,6 +361,8 @@ def initialize(self, batch_info, sleeptime=60):
         self._exec_dag.set_adapter(batch_info)
         self._study.store_metadata()
         self._setup = True
+        # Write initial status.csv immediately after graph construction
+        self._exec_dag.write_status()
```
jwhite242: just need the path argument added here and think we'll be good to go

maestrowf/datastructures/core/executiongraph.py
jwhite242: great, thanks for making those changes!   Looks almost ready to go, just a few more general things things:
* throw in the changelog fragment; see here for details on that: https://github.com/llnl/maestrowf/blob/develop/CONTRIBUTING.md#changelog-fragments 
* bump the dev version in the pyproject.toml

and see the other comment in conductor for one bug to fix

maestrowf/conductor.py
atgrier: In the other use of `write_status()` you have `dag.write_status(os.path.split(pkl_path)[0])`, which substituting `pkl_path` is `dag.write_status(os.path.split(os.path.join(self._pkl_path, "{}.pkl".format(self._study.name)))[0])`. I believe this should be equivalent to just `self._pkl_path`, provided there are no slashes in the `self._study.name`. Is that always true?

jwhite242: should be..  but it's probably clearly to use output_path as the base instead of pkl_path like conductor's 'get_status' does

maestrowf/datastructures/core/executiongraph.py
atgrier: I pushed up a new commit with these changes.

jwhite242 (approved):

PR #480, opened 2026-07-29
Relax hypothesis test timings in CI

jwhite242:
Adds local and ci profiles for hypothesis test cases.  Ci profile extends deadlines to handle higher system variability that was causing sporadic test failures due to default hypothesis deadline being exceeded.


doutriaux1 (approved):
neat!

